PR #36, opened 2015-09-15
This should be a locator, not a located

sorawee:
@jpolitz


awstlaur:
The issue referencing this pull request was closed, which makes me think this pull request wants to be closed as well.

sorawee:
I closed it because I submitted the PR!

blerner:
@sorawee I think this PR can be closed -- this file no longer exists as such, and in cpo-main.js where the code's moved to, it seems like it's now more right than this PR is.  Could you confirm and close?

sorawee:
Yup! (also, I will try to do all stuff I promised to do this weekend. Really apologize for the late)

PR #39, opened 2015-10-15
Add particle as an allowed import.

sstrickl:
This adds the new particle module added in brownplt/pyret-lang#580.


jpolitz:
FYI, this seems to build fine; something is up with our Travis credentials in pull request builds.

PR #49, opened 2016-02-10
Removed dependency on CodeMirror fork and fixed lingering highlighter bugs

belph:
@blerner @jpolitz 
By upgrading to the latest CodeMirror version, this edition is able to use `CodeMirror.prototype.getLineTokens`, which removes the dependency on the quadratic-time `getTokenAt` function.


blerner:
What do we need the `emacs.js` file for?  It's mentioned in our makefile, but seems never to be used by anything relevant to CPO...

src/web/js/codemirror/pyret-mode.js
```diff
@@ -0,0 +1,823 @@
...
+        ls.delimType = pyret_delimiter_type.SUBKEYWORD;
+      }
+    } else if (state.lastToken === "else") {
+      if (hasTop(ls.tokens, "IF")) {
+        if (ls.curOpened.fn > 0) ls.curOpened.fn--;
+        else if (ls.deferedOpened.fn > 0) ls.deferedOpened.fn--;
+        else ls.curClosed.fn++;
+        ls.deferedOpened.fn++;
+        ls.tokens.push("WANTCOLON");
+        ls.subkwParent = ["if"];
```
blerner: Delete this (and line 467, and anywhere else it appears...)

belph:
@blerner and I discussed this in person, but, just to put it on the record:

- `emacs.js` was copied since it was being put in `build` by the `Makefile`. I will remove it later today.
- Line 467 is old code which I should have deleted.

Both of these should be coming out in a new commit later today.

jpolitz:
`emacs.js` came from some Swat students who wanted to try it.  I have little evidence that anyone actually _used_ it.  I'm happy removing it.  Just FYI.

belph:
@blerner @jpolitz This should be good to go now, unless I'm forgetting something.

PR #50, opened 2016-02-20
Adding a Ctrl+I hotkey to reindent all lines (following code from https://groups.google.com/d/msg/codemirror/oGeWPzZynxo/cJoQMzkj0QoJ)

blerner:
This should provide an equivalent experience to WeScheme's reindent-all.  It's bound to Ctrl+I, which is already a hotkey in (at least) Firefox, but as long as the editor has the keyboard focus, it works fine.  Tested on CPO/Pyret-master, and it works for me.

### DIFF
--- a/src/web/js/repl-ui.js
+++ b/src/web/js/repl-ui.js
@@ -35,11 +35,19 @@ define(["js/ffi-helpers", "js/runtime-util", "trove/image-lib", "./check-ui.js",
 
     var useLineNumbers = !options.simpleEditor;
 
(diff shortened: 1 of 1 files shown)


shriram:
How will people find out about it?

blerner:
That's a general question -- all of our keyboard UI is rather undiscoverable.  This particular hotkey is identical to WeScheme, so it should make @schanzer happy that it's now merged, but yes, we certainly need to improve our hotkey help story.

PR #52, opened 2016-02-27
Enable ES6 Harmony for mocha

sorawee:


sorawee:
Joe just fixed the problem

PR #54, opened 2016-03-31
Added the ability to use the editor while the REPL spins up

belph:
**Disclaimer:** This PR falls into the "merge it if you want it" category.

This branch modifies the CPO editor so that the editor can be used while the Pyret REPL spins up in the REPL side window. Don't merge it yet; I just want to know if this is a desired feature. If so, I will further refactor this to load the editor even sooner (before any Pyret runtime is loaded). 

As for the inspiration for this, while I have noticed a marked decrease in bootup time (might just be my machine), I can recall many times during which I have sat waiting for the REPL to spin up and have thought, "Wouldn't it be great if I could be typing while this happens?" As such, this realizes that idea.

Here is a screenshot of what the loading sequence now looks like:
pyret-async


jpolitz:
We very much want this.  There are several kinds of race conditions that can occur during bootup – Drive loading vs. editor loading, run button availability, etc.

Did you have to think about issues like that when making these changes?

jpolitz:
We very much want this.  There are several kinds of race conditions that can occur during bootup – Drive loading vs. editor loading, run button availability, etc.

Did you have to think about issues like that when making these changes?

belph:
I did my best to structure this in the following way:

Anything which can be done immediately (from the original `doWithRepl` function) without other Pyret functions' outputs, do immediately in the new function within the `setTimeout`.

The only other things leftover depend on `repl`, so we make our new asynchronous function's promise resolve to a continuation which "backpatches" what `doWithRepl` needed the REPL for.

I can do some looking around to make sure that none of the other files referenced by the REPL spinup process modify any of the DOM elements referenced here, but I ran it a number of times on my machine with what appeared to be consistent behavior (so that's promising, at least).

(tl;dr I'll look for any more hidden race conditions, but this should be async-safe)

schanzer:
omg yes

jpolitz:
OK, @belph, sounds like you did think through it!  :-)

Have you tested with loading a GDrive file (e.g. with both #share and #program URLs)?

The program loading can race with the editor being ready to have its text set, which can be an issue.

Anyway, we want this, so I'll just review the code carefully at some point, since it sounds like you did the right things to get it in place.

shriram:
OMG. We can't have enough of this. The load time of Pyret is the most immediately, visible terribleness of the language. This will improve that at least in the form of a psychological trick.

belph:
I think that I've put the loading as early into the pipeline as possible at this point.

If everything passes a code review (and no one objects), I'd say it's now safe to merge this PR.

At the same time, I am looking into how we might be able to utilize RequireJS's optimizer (which can integrate Google's Closure Compiler) to reduce the webpage load. Seeing how that might affect interfacing with the Pyret runtime is, of course, a large concern with this idea. Stay tuned.

jpolitz:
@awstlaur did some work on minifying the Pyret standalone, too, so he might have some insight.

Of course, any work on that will need to redone for new module shapes.

blerner:
Philip and I are working through this now.  I'm running it in a VM, which very nicely emulates an underpowered, slow machine...  This PR allows us to start typing into the definitions area before the spinning skull finishes, but there's a big catch.  Right around when we begin initializing the Pyret runtime, *everything* comes to a standstill: the head stops spinning, and text stops being entered into the definitions area, and everything backs up until the browser finishes digesting the huge bolus of code that is Pyret.  Weirdly, we *don't* get an unresponsive-script warning, but nevertheless the browser is unresponsive.  

I *think* we can get a bit better once the modules land, because we'll (or ought to) have the ability to bounce the stack between each module being loaded, and hence give control back to the browser at a bit finer granularity than just "make a runtime".  So, this PR is a good start, but it's not a miracle cure for the slow Pyret startup time.  And, since it gives the illusion of being able to type in the definitions pane and then kinda takes that away from you, it gives a very odd user experience.

Two other (more) minor points: first, the blue Run button is currently blue even though the runtime isn't loaded and the button won't do anything yet, so that's a bit misleading.  And second, the switchover from the [Program Name][More] to [Connect to GDrive][More] commands in the top toolbar should either switch over even sooner, or just not be displayed until we've loaded the google connectivity and determined which version to show.

shriram:
Frankly, I'd rather wait until we can have a seamless experience. It's actually worse to give the impression you can get started and then take it away, than to not give the impression at all. (No criticism of @belph, it's awesome to have this!) Once we can bounce during module load, this should be in just the right place.

jpolitz:
The “catch” sounds like what Pyret currently does, and we can't avoid yet, right?

How long is your browser unresponsive for?  Is it for _longer_ than just loading the usual page?

The Run button activation seems like just another load activity to sync on; that could be useful to do right now.

jpolitz:
I wonder if we could set things up so that the REPL prompt appears as well, and it's just the first click of "Run" and the first "Enter" at the REPL that waits for the runtime to load.

Then the user can load up, look around, type a command, and likely just wait a second or two for the first run.  Sometimes they may not even notice the load.  We could put some gray text in the bottom right to let them know what's going on if they want to track it.

schanzer:
I REALLY like this suggestion. A little "spinning wheel" icon to the right of the cursor or on the right-hand side of the window would be great. Bonus points for blinking it if the user hits enter!

(Then use the same icon on the "Run" button.)

schanzer:
Didn't this already get merged a while back?

jpolitz:
Not exactly this, though there is some load-order stuff done, but in a different way.

There's a collection of minify/babelify/load order things to think through all together.  There are several existing PRs that each attack part of the problem, but it's worth thinking through the solution more comprehensively, since there are things like webpack that are built for this.  They interact with the load order of the page, whether the output of the compiler and related libraries can be in ES6 or pure ES5, and if all of that spans both `pyret-lang` and `code.pyret.org`, or if it's only a CPO thing.

schanzer:
@jpolitz , this functionality has been on CPO for a while now. OK to close?

PR #68, opened 2016-05-06
new plot lib

sorawee:
@jpolitz @blerner FYI


src/web/js/repl-ui.js
```diff
@@ -250,6 +250,33 @@ define(["js/ffi-helpers", "js/runtime-util", "trove/image-lib", "./check-ui.js",
         });
         animationDiv.append(dom);
       });
+    runtime.setParam("d3-port", function(dom, width, height, onExit) {
+        animationDiv = $("<div>").css({"z-index": 10000});
+        output.append(animationDiv);
+        function onClose() {
+          Jsworld.shutdown({ cleanShutdown: true });
```
sorawee: I think this doesn't really need `Jsworld` stuff at all. Will try to get rid of it in next commit

src/web/js/repl-ui.js
```diff
@@ -250,6 +250,33 @@ define(["js/ffi-helpers", "js/runtime-util", "trove/image-lib", "./check-ui.js",
...
+          onExit();
+          showPrompt();
```
sorawee: We don't want to always `showPrompt`. This needs a more sophisticated condition.

PR #72, opened 2016-05-21
Re-engineered the Google API infrastructure to be more extensible with new APIs

belph:
Essentially, this adds a layer between the Drive API (what was in `programs.js`) and the `gapi` object. This layer provides the same loading/API access while abstracting away things like making sure the session is still authenticated with our infrastructure.

This has the effect of removing the need to wrap each API call with a call to `gQ`, along with the benefit of pulling out the boilerplate code needed for new API integrations.

For example, I would like to use a similar infrastructure to interface with the new sheets API, but doing so requires the reintroduction of a significant amount of code. On the other hand, this provides the needed abstractions to avoid a large amount of this rewriting.


src/web/editor.template.html
```diff
@@ -73,6 +73,7 @@
 </div>
 </div>
 <div id="footer"></div>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/es5-shim/4.5.7/es5-shim.min.js"></script>
```
jpolitz: What breaks if we don't include this?  CPO/Pyret assumes an ES5 browser, so I don't want to add extra dependencies if it's avoidable.

src/web/js/google-apis/api-wrapper.js
```diff
@@ -0,0 +1,358 @@
...
+
+// GAPI Client APIs whose methods have been wrapped with calls to gQ
+// (This is kept in the global state to avoid unneeded reloading)
+var _GWRAP_APIS = {};
+
+/**
+ * Base module for interfacing with Google APIs.
+ * Note that any APIs will only be authenticated
+ * with the OAuth scopes found in /src/google-auth.js.
+ * @param {boolean} immediate - See the description of `immediate`
```
jpolitz: Is this just structured docs, or are you using something like closure compiler to check?  Cool if you are, we could work that into the build workflow.

src/web/js/google-apis/api-wrapper.js
```diff
@@ -0,0 +1,358 @@
...
+    function processKey(key) {
+      function processObject(obj, dest) {
+        Object.keys(obj).forEach(function(key) {
+          if (typeof(obj[key]) === "object") {
+            dest[key] = {};
+            processObject(obj[key], dest[key]);
+          } else if (typeof(obj[key]) === "function") {
+            // Use one argument, since it seems that's all that's
+            // used, and there's no `Function.prototype.apply` equivalent
+            // that doesn't change the function's context
```
jpolitz: I don't quite understand this comment.  What do you mean by "the function's context"?

src/web/js/google-apis/drive.js
```diff
@@ -349,40 +278,15 @@ function createProgramCollectionAPI(clientId, apiKey, collectionName, immediate)
...
-  return initialAuth.then(function(_) {
-    var d = Q.defer();
-    gapi.client.load('drive', 'v2', function() {
-      d.resolve(initialize())
-    });
-    return d.promise;
-  });
-  return initialAuth;
+  var ret = Q.defer();
+  gwrap.load({name: 'drive',
```
jpolitz: So the idea is there will be another `gwrap.load()` call for the sheets API, and any other Google APIs we end up using?

Could we load the file picker UI through this?  That's an open feature request (for loading images into the editor) that I'd like to tackle sooner rather than later.

src/web/editor.template.html
belph: Nothing should break. I didn't realize we assumed ES5 compatibility, so I threw it in.

src/web/js/google-apis/api-wrapper.js
belph: I wish I had something that nifty, but they're structured documentation. The inclusion of the Pyret runtime makes using the Closure compiler tricky, from what I remember (I played with the idea a while back)

src/web/js/google-apis/api-wrapper.js
belph: The meaning of `this` within the function body

src/web/js/google-apis/drive.js
belph: That's exactly the idea. I'd imagine that we could integrate the picker API without too much difficulty. I can look into that, if you'd like.

src/web/editor.template.html
belph: Fixed by 3a05fd1

jpolitz:
This seems great overall, especially considering it's silly to not share auth between different G services.  

The travis build says it's missing a file called `api-wrapper.js`:

https://travis-ci.org/brownplt/code.pyret.org/builds/131937949

belph:
Okay, it looks like the Travis build is now fixed. :+1:

src/web/js/google-apis/api-wrapper.js
```diff
@@ -0,0 +1,358 @@
...
+            dest[key] = (function(args, skipAuth) {
+              return gQ(original(args), skipAuth);
+            });
+          } else {
+            dest[key] = obj[key];
+          }
+        });
+      }
+
+      if (Object.keys(_GWRAP_APIS).includes(key)) {
```
blerner: Is this the same as `if (key in _GWRAP_APIS)`?

belph: Yes, it is. I'll change that.

src/web/js/google-apis/api-wrapper.js
```diff
@@ -0,0 +1,358 @@
...
+        // Do nothing
+        return null;
```
blerner: Why null, and not the existing value?

More generally, is it possible that multiple Google APIs will have the same name?  (sheets.load might be the same as files.load, or something)  I'm not sure I fully understand this code yet, but do you have some nesting structure in this `_GWRAP_APIS` cache to avoid problems like that?

belph: I fat-fingered that (I wrote it before opting to return anything at all from this specific function).

In that case, there would be no problem. The way that `gwrap` is set up, the keys in `_GWRAP_APIS` would be `files` and `sheets` (in your example). There should never be a conflict, since we are using the same schema as the `gapi.client` object does (we just wrap everything in the authentication infrastructure)

blerner: Oh, I think I missed the distinction between `processKey` and `processObject` (because the latter function is nested in the former, I kinda missed it).

jpolitz:
Cool, I'm happy with this if we want to push the button.

schanzer:
@ds26gte may be relevant to you

Issue #63, opened 2016-04-13
Open files using Google's Picker API


schanzer:
A user should be able to see all their Pyret files without opening a separate tab and being confronted with the GDrive interface. This is especially valuable for schools, where teachers may want to avoid showing students GDrive altogether.

PR #73, opened 2016-05-24
Added Google Picker UI Support

### LINKED ISSUE #62: Open/Include images using Google's Picker API
Originally opened at https://github.com/brownplt/pyret-lang/issues/546

An example of a working Image picker can be found in WeScheme, at https://github.com/bootstrapworld/wescheme2012/blob/master/war-src/js/openEditor/editor.js#L571-L659

belph:
This Pull Request adds support for the Google Picker UI.

Users can open both Pyret files and Images with the interface.

This PR fixes #62 and #63.

### DIFF
--- a/src/web/js/google-apis/picker.js
+++ b/src/web/js/google-apis/picker.js
@@ -1,0 +1,128 @@
+// Loosely adapted from https://gist.github.com/Daniel15/5994054
+
+/**
+ * Creates the Google File Picker API
+ */
+function FilePicker(options) {
+  options = options || {};
+
+  this.dataHandler = options.onSelect;
+  this.onLoaded = options.onLoaded || function(){};
+  this.onError = options.onError || function(m) { throw new gwrap.GAPIError(m); };
+  this.onInternalError = options.onInternalError || this.onError;
+
+  function onPickerLoaded(drive) {
+    return function() {
+      this.init(drive, google.picker);
+      this.onLoaded();
+    };
+  }
+
+  storageAPI.then((function(drive){
+    google.load('picker', '1', { 'callback': onPickerLoaded(drive).bind(this) });
+  }).bind(this));
+}
+
+FilePicker.prototype.init = function(drive, picker) {
+  this.open = this.initOpen(drive, picker);
+  console.log("Initializing File Picker");
+  // this.openOn does not need initialization
+};
+
+FilePicker.prototype.initOpen = function(drive, picker) {
+
+  var validateSelection = (function(sel) {
+    if (sel.length === 0) {
+      // Should be impossible, since we only run this if the user
+      // has picked something (see pickerCallback)
+      this.onInternalError("Internal Error (please report to developers): "
+                           + "Should be impossible: "
+                           + "Picker selection length is zero");
+    } else {
+      var firstType = sel[0][picker.Document.TYPE];
+      // Check that all selected files are of the same type
+      sel.forEach((function(f) {
+        if (f[picker.Document.TYPE] !== firstType) {
+          this.onError("Cannot open files of multiple types "
+                       + "(Expected: " + firstType
+                       + "; Received: " + f[picker.Document.TYPE]
+                       + ")");
+        }
+      }).bind(this));
+      switch (sel[0][picker.Document.TYPE]) {
+      case "file": // Pyret file
+        break;
+      case picker.Type.PHOTO: // Photo
+        // Fix photo permissions
+        var permissions = {
+          'role': 'reader',
+          'type': 'anyone',
+          'value': 'default',
+          'withLink': true
+        };
+        sel.forEach((function(photo) {
+          var fileId = photo[picker.Document.ID];
+          gwrap.drive.permissions.insert({
+            'fileId': fileId,
+            'resource': permissions
+          }).catch(this.onInternalError);
+        }).bind(this));
+        break;
+      default:
+        this.onError("Cannot open file of unknown type: " + sel[0][picker.Document.TYPE]);
+      }
+    }
+  }).bind(this);
+
+  var pickerCallback = (function(data) {
+    if (data[picker.Response.ACTION] == picker.Action.PICKED) {
+      var selection = data[picker.Response.DOCUMENTS];
+      validateSelection(selection);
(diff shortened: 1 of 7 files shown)


src/web/js/google-apis/picker.js
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  storageAPI.then((function(drive){
+    google.load('picker', '1', { 'callback': onPickerLoaded(drive).bind(this) });
+  }).bind(this));
+}
+
+FilePicker.prototype.init = function(drive, picker) {
+  this.open = this.initOpen(drive, picker);
+  console.log("Initializing File Picker");
```
blerner: Probably shouldn't leave `console.log`s in place in final code?

src/web/js/google-apis/picker.js
```diff
@@ -0,0 +1,128 @@
...
+     * A Picker View which displays Pyret files which users may load.
+     */
+    var pyretView = new picker.DocsView(picker.ViewId.DOCUMENTS);
+
+    // [Image picker code is based on the Picker code from WeScheme:
+    // https://github.com/bootstrapworld/wescheme2012/blob/5b03bac247ac182c356d73fc3e7283ef1086777f/war-src/js/openEditor/editor.js#L571-L659 ]
+    /**
+     * A Picker View which displays images users may load.
+     */
+    var imageView = new picker.View(picker.ViewId.DOCS);
```
blerner: `DOCS` here vs `DOCUMENTS` above?  Is that deliberate?

src/web/js/google-apis/picker.js
```diff
@@ -0,0 +1,128 @@
...
+    imageView.setMimeTypes("image/png,image/jpeg,image/jpg,image/gif");
+
+    var buildInstance = (function(parentId) {
+      pyretView.setParent(parentId);
+
+      this.pickerInstance = new picker.PickerBuilder()
+        //.enableFeature(picker.Feature.NAV_HIDDEN)
+        .enableFeature(picker.Feature.MULTISELECT_ENABLED)
+        .setTitle("Select a Pyret File or Image")
```
blerner: Probably should pass in a title string here, rather than have an ambivalent prompt instead...

src/web/js/google-apis/picker.js
belph: Yes; `DOCUMENTS` limits the view to text documents, while `DOCS` allows for any file type (i.e. images as well)

src/web/js/google-apis/picker.js
belph: Probably right

src/web/js/google-apis/picker.js
schanzer: I believe there's a way to specify a fixed list of mime types. People store all KINDS of things in their  GDrives, and it would be great to reduce the likelyhood of a student or user hitting the wrong thing and getting an error message because they tried to load some bizarro filetype as a bitmap.

src/web/js/google-apis/picker.js
belph: Not sure if that's fitting with the way it's currently engineered. Right now, the picker UI can't really do anything other than select the files and images (which is hard-coded a few lines above this), so I don't see how parametrizing the title is really needed. Perhaps if the library was more general purpose that would make sense, but I am not sold.

schanzer: Spoke too soon - looks like you've got this handled already. ;)

blerner: Many of the messages we've presented to students before have said "Try doing this thing or that thing to fix problem X".  But since you're using this in the context of either picking files or picking images, seems like you should be able to pass in the string to specialize the prompt, so that it's clearer what action should be taken right now.

src/web/js/google-apis/picker.js
belph: @jpolitz and I were discussing this -- currently, Pyret files are stored with the `text/plain` MIME type (and, yes, now that I think about it, we should probably limit the view to that type), so there is no MIME-filtering way to only show Pyret files (we do, however, start the Pyret file view in the Pyret save folder, which should usually only contain Pyret files). The implications of a Pyret-specific MIME type need to be looked into, but until then, this will have to do.

@schanzer, your comment sounds like it is in reference to images specifically. A few lines down, I have the MIME-filtering enabled for image files. Thus, there shouldn't be any way to open non-images as images (unless one is deliberately trying).

src/web/js/google-apis/picker.js
belph: @blerner I think I see what you're saying...and, IIUC, you're mistaken. Files and images are opened using different tabs in the *same* Picker UI. The title is visible on both, since they are the same window (once again, **different tabs**, but same window).

belph:
A demonstration can be found here: http://pyret-picker-ui.herokuapp.com/editor

jpolitz:
This is so. Freaking. Cool.  OMG.

@schanzer this is the version you want to check out (click the Open Button).

How is it finding Pyret files?  Just by filtering on things created by "this app"?

jpolitz:
Oh I see the discussion above now, sorry.  This does seem to have filtered things by recently created Pyret files when I open it, though.

jpolitz:
Also, that's a pretty intimidating permissions screen.

screen shot 2016-05-24 at 7 04 46 pm

belph:
@jpolitz It might look like it's only recently created files because, IIUC, it is not difficult to trick Google into thinking you're a "new" Pyret and creating a new collections folder. This has the effect of only putting the files created since the most recent folder was made in view.

Thankfully, this is only something we notice since we develop CPO. End-users should have no such problems.

Regarding the permissions, yes, they are...large. However, I am relatively certain that only permissions in use are being requested...

schanzer:
Yup - this is one of the highest-priority UI items for me. @belph , you're my hero! 

Now, of course I'm chomping at the bit to get this into CPO, and ultimately into spyret...

jpolitz:
Oh! @schanzer I totally missed that this had (boring) conflicts that failed a clean merge.  Just pushed this.

schanzer:
THERE it is. Gorgeous!!!

PR #80, opened 2016-06-16
Beginnings of build system refactoring using webpack

pcardune:
Throwing this up here for a first look. Some things that I added include:

- javascript linting using [eslint](http://eslint.org/)
- basic [webpack](http://webpack.github.io/docs/) configuration
- minor adjustments to the Makefile to work with webpack
- minor adjustments to `beforePyret.js` that shows a little bit of what webpack can do.

webpack is a quite excellent javascript build system that can do all the bundling, minifying, etc. etc. you expect from a build system. I've used to great effect on a lot of different projects.

### DIFF
--- a/src/web/editor.template.html
+++ b/src/web/editor.template.html
@@ -104,7 +104,6 @@
 <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.10.2/jquery.min.js"></script>
 <script src="https://ajax.googleapis.com/ajax/libs/jqueryui/1.10.4/jquery-ui.min.js"></script>
 <script src="/js/loader.js"></script>
-<script src="/js/url.js"></script>
 <script src="/js/colorspaces.js"></script>
 <script src="/js/q.js"></script>
 <script src="/js/codemirror.js"></script>
--- a/src/web/editor.template.html
+++ b/src/web/editor.template.html
@@ -132,11 +132,6 @@ define("seedrandom", [], function() {
 });
 </script>
 <script src="/js/share.js"></script>
-<script>
-var shareAPI = makeShareAPI("{{{ CURRENT_PYRET_RELEASE }}}");
-var makeShareLink = shareAPI.makeShareLink;
-var makeHoverMenu = shareAPI.makeHoverMenu;
-</script>
 <script src="/js/google-apis/api-wrapper.js"></script>
 <script src="/js/google-apis/drive.js"></script>
 <script src="/js/google-apis/picker.js"></script>
@@ -156,7 +151,6 @@ function handleGoogLoad() {
 <script>console.log(window.performance.now())</script>
 <script src="js/beforePyret.js"></script>
 <script>var APP_BASE_URL = "{{{ BASE_URL }}}";</script>
-<script>var PYRETSRC = "{{{ PYRET }}}"</script>
 
 <!-- <script type="text/javascript"
         src="https://caja.appspot.com/caja.js"></script> -->
--- a/src/web/js/beforePyret.js
+++ b/src/web/js/beforePyret.js
@@ -1,4 +1,6 @@
-/* global $ jQuery CPO CodeMirror storageAPI Q createProgramCollectionAPI makeShareLink */
+/* global $ jQuery CPO CodeMirror storageAPI Q createProgramCollectionAPI makeShareAPI */
+
+var shareAPI = makeShareAPI(process.env.CURRENT_PYRET_RELEASE);
 
 var url = require('url.js');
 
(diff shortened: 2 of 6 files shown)


webpack.config.js
```diff
@@ -0,0 +1,28 @@
...
+  },
+  devtool: IS_PRODUCTION ? 'source-map' : 'inline-source-map',
+  entry: {
+    "beforePyret": './src/web/js/beforePyret.js',
+  },
+  plugins: [
+    new webpack.DefinePlugin({
+      'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV),
+      'process.env.PYRET': JSON.stringify(process.env.PYRET),
+      'process.env.CURRENT_PYRET_RELEASE': JSON.stringify(process.env.CURRENT_PYRET_RELEASE),
```
jpolitz: Nice.  This is an improvement over the current templating, I think.

Where does the `process` sorcery happen?  I grepped the `build` directory for `PYRET` and don't see where it's set.

src/web/js/beforePyret.js
```diff
@@ -1,48 +1,55 @@
...
+window.ct_error = function(/* varargs */) {
   if (window.console && LOG) {
     console.error.apply(console, arguments);
   }
-}
+};
 var initialParams = url.parse(document.location.href);
 var params = url.parse("/?" + initialParams["hash"]);
-var highlightMode = "mcmh";
+var highlightMode = "mcmh"; // what is this for?
```
jpolitz: @jswrenn ping (I know the answer—it corresponds to the highlighting choices in the dropdown—but Jack can probably answer follow-on questions better)

src/web/editor.template.html
```diff
@@ -104,7 +104,6 @@
 <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.10.2/jquery.min.js"></script>
 <script src="https://ajax.googleapis.com/ajax/libs/jqueryui/1.10.4/jquery-ui.min.js"></script>
 <script src="/js/loader.js"></script>
-<script src="/js/url.js"></script>
```
jpolitz: So the plan would be to give all/as may as possible of these these modules similar treatment to url, right?

webpack.config.js
```diff
@@ -0,0 +1,28 @@
...
+    }),
+  ].concat(IS_PRODUCTION ? [
```
jpolitz: What should I expect if I put `NODE_ENV="production"` in my `.env` and rebuild (after a `make clean`)?

I get "The page has a redirect loop" after restarting the server and reloading the page, but I'm not sure if I'm missing anything/etc.

webpack.config.js
pcardune: It actually gets inlined, so you won't find `PYRET` in the build directory. Rather you will find whatever the value of `process.env.PYRET` was at build time.

This is especially useful when combined with dead code removal because you can do stuff like this:

```javascript
if (process.env.DEBUG) {
  console.log("this won't show up in production/minified js files");
}
```

and it gets turned into this:

```javascript
if (undefined) {
  console.log("this won't show up in production/minified js files");
}
```

which the static analyzer in the uglify plugin will recognize as dead code and completely remove.

src/web/editor.template.html
pcardune: Yes. For that reason, it would be helpful to know what the implicit dependency tree is for these script tags. I'm slightly terrified to remove something here that happens to be used somewhere in the bowels of the pyret runtime, resulting in breaking everything.

The goal is to remove all globals and make all the module dependencies explicit.

webpack.config.js
pcardune: you know, I haven't actually tested out the production config yet. But you should get:
- minification of the generated js file
- production quality source maps

I'll take a look and see if I can figure out what that issue is.

pcardune:
as for the travis-ci build issues, it's complaining about sauce labs authentication. Do you know how I can fix that?

jpolitz:
If you make this a branch on the main repo (or just push to horizon), the travis build will kick in.  Travis just doesn't include encrypted environment variables (https://github.com/brownplt/code.pyret.org/blob/horizon/.travis.yml#L18) on PRs from external repos.  I don't see an obvious way to set that up to work on an external repo's PR.

It is probably straightforward, if a bit time-consuming, to set up a Sauce account and have the tests run from there (it's free for open source projects), but it's also easy to make a branch on this repo and do it that way, too.

src/web/editor.template.html
jpolitz: There isn't anything here that `pyret-lang` depends on to run the compiler/language runtime.  Actually, since `world` and `image` are now in the CPO repo as well, the situation is actually better than it was, dependency-wise.

So any uses of globals defined in these script tags will be within the CPO repository, not lurking elsewhere, at least.  I'll write up more dependency information in a comment on the PR as a whole.

webpack.config.js
jpolitz: Ah, I see, and webpack puts some excitement at the bottom of `beforePyret.js` as well.  Cool, I know more where to look to understand that layout now.

PR #82, opened 2016-06-16
Beginnings of an updated build system

pcardune:
Throwing this up here for a first look. Some things that I added include:

- javascript linting using [eslint](http://eslint.org/)
- basic [webpack](http://webpack.github.io/docs/) configuration
- minor adjustments to the Makefile to work with webpack
- minor adjustments to `beforePyret.js` that shows a little bit of what webpack can do.

webpack is a quite excellent javascript build system that can do all the bundling, minifying, etc. etc. you expect from a build system. I've used to great effect on a lot of different projects.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,6 +7,7 @@ REDISCLOUD_URL=""
 USE_STANDALONE_PYRET=true
 PYRET_RELEASE_BASE="/js"
 CURRENT_PYRET_RELEASE=""
+PYRET="http://localhost:5000/js/cpo-main.jarr"
 PORT=5000
 NODE_ENV=development
 OK_GOOGLE_IDS="{\"0B32bNEogmncO\":\"joe.politz@gmail.com\"}"
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,21 @@
+{
+  "parser": "babel-eslint",
+  "rules": {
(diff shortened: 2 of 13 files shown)


pcardune:
@jpolitz looks like these builds failed too, but checking out the travis ouput, I can't tell what the problem is. It looks like it just timed out running one of the build steps? Any idea what that means?

pcardune:
@jpolitz also, I was able to repro the infinite redirect when `NODE_ENV=production`. It appears that `server.js` contains redirect code that only runs in production which tries to redirect you to `https`. The browser sees that you can't connect to ssl (since you aren't running an ssl server locally) and drops you back to http, which then begins the cycle anew. This definitely doesn't have anything to do with the code in this pull request.

jpolitz:
Oh, duh on the redirect.

The build failed because a global (`makeHoverMenu`) went away; the timeout error is selenium waiting for the editor to load and it never actually loaded.  I pushed the fix for that.

The test just waits for the loader gif to be hidden, and doesn't understand errors very well.  The output is here:

https://saucelabs.com/beta/tests/53952c86aab046dd8e7a13a678792004/commands

But you don't have a good way to get to that.  I will try to see if there's a good way to get the test framework to print out the tests it tries; looking at the screenshot is often a good way to find out what went wrong.

PR #91, opened 2016-06-30
Update to new api for interfacing with pyret-ide

pcardune:
This stubs out the api for making pyret ide interface with the pyret runtime.

### DIFF
--- a/src/web/js/cpo-ide-hooks.js
+++ b/src/web/js/cpo-ide-hooks.js
@@ -93,6 +93,13 @@
     });
     var builtinsForPyret = runtime.ffi.makeList(builtins);
 
+    function parse(source, uri) {
+      var parse = runtime.getField(runtime.getField(parsePyret, "values"), "surface-parse")
+      return runtime.safeTail(function() {
+        return parse.app(source, uri);
+      });
+    }
+
     function compile(ast) {
       var compileAst = gmf(cpo, "compile-ast");
       return runtime.safeTail(function() {
@@ -109,7 +116,7 @@
 
     window.CPOIDEHooks = {
       runtime: runtime,
-      parsePyret: parsePyret,
+      parse: parse,
       compile: compile,
       run: run
     };
--- a/src/web/js/cpo-ide-hooks.js
+++ b/src/web/js/cpo-ide-hooks.js
@@ -23,9 +23,6 @@
       name: "load-lib"
     },
     { "import-type": "builtin",
-      name: "builtin-modules"
-    },
-    { "import-type": "builtin",
       name: "cpo-builtins"
     },
     {
@@ -39,86 +36,18 @@
   provides: {},
   theModule: function(runtime, namespace, uri,
                       compileLib, compileStructs, pyRepl, cpo,
-                      runtimeLib, loadLib, builtinModules, cpoBuiltins, parsePyret,
+                      runtimeLib, loadLib, cpoBuiltins, parsePyret,
                       cpoModules
   ) {
-
-    var gf = runtime.getField;
-    var gmf = function(m, f) { return gf(gf(m, "values"), f); };
-
-    function findModule(contextIgnored, dependency) {
-      // TODO(joe): enhance this with gdrive locators, etc, later
-      return runtime.safeCall(function() {
-        return runtime.ffi.cases(gmf(compileStructs, "is-Dependency"), "Dependency", dependency,
-          {
-            builtin: function(name) {
-              var raw = cpoModules.getBuiltinLoadableName(runtime, name);
-              if(!raw) {
-                throw runtime.throwMessageException("Unknown module: " + name);
-              }
-              else {
-                return gmf(cpo, "make-builtin-js-locator").app(name, raw);
-              }
-            },
-            dependency: function(protocol, args) {
-              console.error("Unknown import: ", dependency);
-            }
-          });
-       }, function(l) {
-          return gmf(compileLib, "located").app(l, runtime.nothing);
-       }, "findModule");
-    }
-    var pyFindModule = runtime.makeFunction(findModule, "find-module");
-
-    // NOTE(joe): This line is "cheating" by mixing runtime levels, and uses
-    // the same runtime for the compiler and running code.  Usually you can
-    // only get a new pyret-viewable Runtime by calling create, but here we
-    // magic the current runtime into one.
-    // Someday Pyret will be quick enough that we won't need to save theses
-    // seconds of instantiation.
-    var pyRuntime = gf(gf(runtimeLib, "internal").brandRuntime, "brand").app(
-      runtime.makeObject({
-        "runtime": runtime.makeOpaque(runtime)
-      }));
-    var pyRealm = gf(loadLib, "internal").makeRealm(cpoModules.getRealm());
-
-    var builtins = [];
-    Object.keys(runtime.getParam("staticModules")).forEach(function(k) {
-      if(k.indexOf("builtin://") === 0) {
-        builtins.push(runtime.makeObject({
-          uri: k,
-          raw: cpoModules.getBuiltinLoadable(runtime, k)
-        }));
-      }
-    });
-    var builtinsForPyret = runtime.ffi.makeList(builtins);
-
-    function parse(source, uri) {
-      var parse = runtime.getField(runtime.getField(parsePyret, "values"), "surface-parse")
-      return runtime.safeTail(function() {
-        return parse.app(source, uri);
-      });
-    }
-
-    function compile(ast) {
-      var compileAst = gmf(cpo, "compile-ast");
-      return runtime.safeTail(function() {
-        return compileAst.app(ast, pyRuntime, pyFindModule, gmf(compileStructs, "default-compile-options"));
-      });
-    }
-
-    function run(jsSrc) {
-      var run = gmf(cpo, "run");
-      return runtime.safeTail(function() {
-        return run.app(pyRuntime, pyRealm, jsSrc);
-      });
-    }
-
     window.CPOIDEHooks = {
       runtime: runtime,
-      parse: parse,
-      compile: compile,
-      run: run
+      cpo: cpo,
+      parsePyret: parsePyret,
+      loadLib: loadLib,
+      runtimeLib: runtimeLib,
+      compileStructs: compileStructs,
+      compileLib: compileLib,
+      cpoModules: cpoModules
     };
     return runtime.makeModuleReturn({}, {});
   }
--- a/src/web/js/cpo-ide-hooks.js
+++ b/src/web/js/cpo-ide-hooks.js
@@ -42,20 +42,76 @@
                       runtimeLib, loadLib, builtinModules, cpoBuiltins, parsePyret,
                       cpoModules
   ) {
+
+    var gf = runtime.getField;
+    var gmf = function(m, f) { return gf(gf(m, "values"), f); };
+
+    function findModule(contextIgnored, dependency) {
+      // TODO(joe): enhance this with gdrive locators, etc, later
+      return runtime.safeCall(function() {
+        return runtime.ffi.cases(gmf(compileStructs, "is-Dependency"), "Dependency", dependency,
+          {
+            builtin: function(name) {
+              var raw = cpoModules.getBuiltinLoadableName(runtime, name);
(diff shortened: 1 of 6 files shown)


src/web/js/ide.js
```diff
@@ -10,11 +8,45 @@ function loadScriptUrl(url) {
...
+const runtimeApi = {
+  parse(src) {
+    return new Promise((resolve, reject) => reject(new Error("Not Implemented")));
+  },
+  compile(ast) {
+    return new Promise((resolve, reject) => reject(new Error("Not Implemented")));
+  },
+  execute(bytecode) {
+    return new Promise((resolve, reject) => reject(new Error("Not Implemented")));
+  },
```
pcardune: I just need some help figuring out how to fill these in using the APIs available on window.CPOIDEHooks (which are defined in cpo-ide-hooks.js)

jpolitz:
Sure thing, started sketching this out.

4.  Is there any code that calls this yet?  I'm thinking about how to test whatever I start adding.
1.  One thing this is missing is way to talk about folding a series of REPL interactions over a context.  It's probably worth defining a JS-facing `PyretContext` datatype that is initially opaque to JS (we can add reflective methods later if we want).  An extra method called `freshContext` can return a clean slate context, which is returned in updated form after `execute`, to be used in future calls.  Then the client of `runtimeApi` can use `freshContext` when starting from the definitions window, and track a running context during a series of interactions.
2.  That said, this is a great place to start; it just won't really have a REPL to speak of.
3.  I suspect the interface also wants `stop`, to play with the stop button.

jpolitz:
Also, looking at this helped clarify for me that the most interesting part of this interface is going to be three components of the return type of execute:

- What a Pyret value looks like (already discussed in some detail)
- What various Pyret _error_s looks like
- What a set of check block results looks like

(The second case—errors—covers the possible errors from `parse` and `compile`.)

Returning strings will get us off the ground, but not much further.  Reporting errors is probably the most interesting part of CPO.

pcardune:
To answer your questions:
1. yes there is. If you run code.pyret.org and go to http://localhost:5000/ide it will execute this code. You can also clone the pcardune/pyret-ide repo, follow the instructions in the README.md file and see a "stub" compiler in action. All it does is basic arithmetic with infix notation, but it properly illustrates the async parse/compile/execute pipeline.
2. yes, I haven't started working on the REPL piece yet. It's a good question who should be responsible for maintain the context state: the runtime or the IDE. I was originally thinking that the runtime would maintain is, and the api would provide a `clearState()` function to clear it. But now that I think about it a bit more, it might make sense for the IDE to keep track of the state (or at least a handle for it) and pass that explicitly into the runtime, so you could potentially have multiple REPLs with multiple different contexts running in the IDE at once.
4. yep, stop is coming next. There is actually a diff out already for pyret-ide that does that part.

Strings is going to be good enough for the moment, but yeah, the really interesting part is going to be those return values! Very excited to see that get fleshed out.

But if this code looks good to you, I'll go ahead and merge it, and then we can replace those not implemented errors with something proper in the next diff.

jpolitz:
OK, this should be a good simple version to start with, returning strings for both values and errors in all cases (except for ASTs, which are a Pyret datatype).  Happy to have this merged.

src/web/js/cpo-ide-hooks.js
```diff
@@ -35,22 +39,79 @@
...
+          raw: cpoModules.getBuiltinLoadable(runtime, k)
+        }));
+      }
+    });
+    var builtinsForPyret = runtime.ffi.makeList(builtins);
+
+    function compile(ast) {
+      var compileAst = gmf(cpo, "compile-ast");
+      return runtime.safeTail(function() {
+        return compileAst.app(ast, pyRuntime, pyFindModule, gmf(compileStructs, "default-compile-options"));
```
blerner: Should I be worried slightly that the `default-compile-options` are hardcoded here?

src/web/js/ide.js
```diff
@@ -10,11 +8,155 @@ function loadScriptUrl(url) {
...
+  _getVal(mod, name) { // TODO(joe): is this an OK place to put helpers for this?
+    var runtime = window.CPOIDEHooks.runtime;
+    return runtime.getField(runtime.getField(window.CPOIDEHooks[mod], "values"), name);
+  },
+  _toReprOrDie(runtime, value, resolve, reject) {
+    runtime.runThunk(
+      () => runtime.toRepr(value),
+      (renderResult) => {
+        if(runtime.isSuccessResult(renderResult)) {
+          reject(renderResult.result);
```
blerner: Er, `resolve`?

src/web/js/ide.js
```diff
@@ -10,11 +8,155 @@ function loadScriptUrl(url) {
...
+    return new Promise((resolve, reject) => {
+      runtime.runThunk(
+        () => window.CPOIDEHooks.compile(ast),
+        (compileResult) => {
+          // NOTE(joe): success here just means the compiler didn't blow up; the result
+          // is a Pyret Either indicating compile errors or a final JS program to run
+          if(runtime.isSuccessResult(compileResult)) {
+            var maybeJS = compileResult.result;
+            if(runtime.ffi.isLeft(maybeJS)) {
+              self._toReprOrDie(runtime, maybeJS, reject)
```
blerner: missing a 3rd `resolve` argument?

src/web/js/ide.js
```diff
@@ -10,11 +8,155 @@ function loadScriptUrl(url) {
...
+
+          // Just doing a barebones dive to retrieve and return the toRepr of
+          // the final value for now, but there are lots of juicy things on
+          // this result, and it's something we should build out an API for.
+          var innerResult = runResult.result.val.result;
+          if(runtime.isSuccessResult(innerResult)) {
+            self._toReprOrDie(runtime, get(innerResult.result, "answer"), resolve, reject)
+          }
+          else {
+            self._toReprOrDie(runtime, innerResult.exn.exn, reject, reject);
```
blerner: Is `reject, reject` deliberate?

src/web/js/ide.js
```diff
@@ -31,32 +48,19 @@ const runtimeApi = {
     // TODO(joe): pass in a URL to uniquely identify this program
     if(!url) { url = "definitions://"; }
     var runtime = window.CPOIDEHooks.runtime;
-    function getVal(mod, name) {
-      return runtime.getField(runtime.getField(mod, "values"), name);
-    }
-    var parse = getVal(window.CPOIDEHooks.parsePyret, "surface-parse");
+    // TODO(joe): is "this" trustworthy here?
+    var self = this;
+    var parse = this._getVal("parsePyret", "surface-parse");
```
pcardune: I think you're better off making `_getVal` a module-level function that doesn't get exported. Then you don't need to worry about `this` and you don't need to worry about anyone monkeying around with `_getVal`.

src/web/js/ide.js
```diff
@@ -31,32 +48,19 @@ const runtimeApi = {
...
       runtime.runThunk(
         () => parse.app(src, url),
         (parseResult) => {
           if(runtime.isSuccessResult(parseResult)) {
             resolve(parseResult.result);
           }
           else {
-            handleError(parseResult);
+            // NOTE(joe): intentionally passing reject twice; want to report an error either way
+            self._toReprOrDie(runtime, parseResult.exn.exn, reject, reject);
```
pcardune: why not just pass a single `reject` and have `_toReprOrDie` only accept a single `reject` seeing as it never calls `resolve()`.

blerner: I think it's supposed to be actually calling `resolve`, rather than always failing, unless I'm misunderstanding something...

src/web/js/ide.js
```diff
@@ -79,7 +106,28 @@ const runtimeApi = {
              (TODO: return richer values for error returns)
   */
   execute(bytecode) {
-    return new Promise((resolve, reject) => reject(new Error("Not Implemented")));
+    var runtime = window.CPOIDEHooks.runtime;
```
pcardune: rather than proliferating references to global `window` properties, I think it would be be preferable to a narrow "window" where we hook up to this global. e.g. something like this:

```javascript
function createRuntimeApi(CPOIDEHooks) {
  return {
    parse() { ... },
    compile() { ... },
    execute() {
      var runtime = CPOIDEHooks.runtime;
      ...
    },
  };
}
```
And then on line 150, you'd have:
```javascript
           resolve(createRuntimeApi(window.CPOIDEHooks));
```

src/web/js/ide.js
jpolitz: Yeah Ben pointed out my stupid.  It should call resolve in the success case and reject in the failure case, but when rendering an error just get two rejects.  I'll fix.

src/web/js/ide.js
jpolitz: Here, yes, because the run failed, so it should be a reject either way.

src/web/js/ide.js
jpolitz: Derp, yeah.

src/web/js/ide.js
jpolitz: Yup.

src/web/js/ide.js
jpolitz: Yup, good idea. That also avoids my motivation for having methods, since it gives us a scope to define local helpers.

src/web/js/cpo-ide-hooks.js
jpolitz: Thanks!  That should have been another TODO comment on `compile` requesting another parameter, much like `url`.

jpolitz:
@pcardune is `pyret-ide` the right place to add tests for this, or did you have something in mind for that in this repo?

I unsurprisingly made stupid mistakes because I was just testing by running the code at `ide` and checking console output/debugger stuff.

jpolitz:
A question about the model for these promises – I picked, without thinking much, that `resolve` would mean "making progress towards getting an answer and successful run" and `reject` meant otherwise.

From the point of view of the IDE, though, getting a parse error or a compiler error or a runtime error isn't really a _failure_ of the API, since those are results that ought to be handled in interesting ways (in fact, they are arguably _more_ important, in a teaching IDE, than the success cases).  Should we reserve `reject` for true failures where something went wrong in the underlying system—bugs—rather than for results that represent errors in the program?

pcardune:
Regarding your question about promises, you make a good point. I suppose each stage of the parse/compile/execute pipeline could resolve some kind of result object that includes a success flag.

Though the nice thing about the current setup is that reject clearly indicates that you should not attempt to continue to the next stage of the pipeline, which would be the case for both bugs and user generated syntax errors.

I feel like the differences between these two options are subtle and therefore I'll ere on the side of sticking with what you've already written. We can sleep on it for a bit until the obvious answer presents itself.

pcardune:
As far as tests are concerned, the best place to test this code is in code.pyret.org. It looks like there isn't currently a unit testing setup in this repo, but we should add one. Unless you have strong preferences about unit testing tools, I can throw up another diff to add the karma test runner with jasmine or mocha to code.pyret.org.

src/web/js/cpo-ide-hooks.js
```diff
@@ -35,22 +39,86 @@
...
-      pyRepl: pyRepl,
-      cpo: cpo,
-      runtimeLib: runtimeLib,
-      loadLib: loadLib,
-      builtinModules: builtinModules,
-      cpoBuiltins: cpoBuiltins,
-      cpoModules: cpoModules,
+      parse: parse,
+      compile: compile,
+      run: run
```
pcardune: Is there any reason not to take the parse/compile/run functions you've created here and just move them up into ide.js? 

The main benefit of moving this logic up the stack is that it becomes easier to change without going through the pyret build process, which at least at this particular moment is a bit more cumbersome than webpack :)

jpolitz:
Yeah, the only tests are run by mocha, and are integration tests; if you run `mocha`, it looks in `test/` and tries to start up selenium webdriver, etc.

I don't know what the best way is to fire up `ide.js` and make calls to it from a test file, so if you want to add some way to do that, it would be great.  It's up to you how what tools/commands to use to run the tests, I don't mind at all they're different from the current `mocha` setup.

src/web/js/cpo-ide-hooks.js
jpolitz: Good idea, done.

PR #93, opened 2016-07-11
New world-lib CPO merge

A-lxe:
Brings @sorawee's world-lib patches (see: https://github.com/brownplt/pyret-lang/pull/732/files?diff=unified) to CPO instead of pyret-lang.

### DIFF
--- a/src/web/js/trove/world-lib.js
+++ b/src/web/js/trove/world-lib.js
@@ -124,6 +124,10 @@
         eventDetachers = null;
       }
       changingWorld.pop();
+
+      if(runningBigBangs.length > 0) {
+        runningBigBangs[runningBigBangs.length - 1].restart();
+      }
     }
 
 
@@ -143,7 +147,7 @@
       clear_running_state();
     };
 
-    // Close all world computations.
+    // Closes the most recent world computation.
     Jsworld.shutdownSingle = function(options) {
       if(runningBigBangs.length > 0) {
         var currentRecord = runningBigBangs.pop();
@@ -156,9 +160,6 @@
         }
       }
       resume_running_state();
-      if(runningBigBangs.length > 0) {
-        runningBigBangs[runningBigBangs.length - 1].start();
-      }
     };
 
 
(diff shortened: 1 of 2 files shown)


src/web/js/trove/world-lib.js
```diff
@@ -97,6 +143,23 @@
       clear_running_state();
     };
 
+    // Close all world computations.
```
sorawee: Please see https://github.com/brownplt/pyret-lang/pull/732#discussion_r66707667

src/web/js/trove/world-lib.js
```diff
@@ -605,8 +668,13 @@
       this.fail = fail;
     }
 
-    BigBangRecord.prototype.restart = function() {
-      bigBang(this.top, this.world, this.handlerCreators, this.attribs);
+    BigBangRecord.prototype.start = function() {
```
sorawee: Ditto.

jpolitz:
@sorawee do you expect this program to count all the way to 100 with this change?

```
include world
include image
import image-structs as IS

big-bang(0, [list: 
    on-tick-n(lam(n):
        if num-modulo(n, 10) == 1 block:
          big-bang(n,
            [list:
              on-tick-n(lam(k): k - 1 end, 0.5),
              stop-when(lam(k): k < 0 end),
              to-draw(lam(k): text(num-to-string(k), 20, IS.blue) end)
            ])
          n + 1
        else:
          n + 1 
        end
      end, 0.5),
    to-draw(lam(n): text(num-to-string(n), 20, IS.violet-red) end),
    stop-when(lam(n): n > 100 end)])
```

jpolitz:
Ah I needed `close-when-stop` I guess.  But if I close manually, it still kills all worlds.  I'm remembering this now.

jpolitz:
I'm going to merge this, because it's definitely an improvement.  Still some thoughts and tests to hammer on, I think.

sorawee:
> But if I close manually, it still kills all worlds.

I thought this patch is supposed to fix that. Something is wrong here. Will take a look.

sorawee:
@jpolitz I think it's not a bug. You just didn't merge https://github.com/brownplt/code.pyret.org/pull/77 which is supposed to be merged along this patch, so the fix is not complete.

PR #105, opened 2016-08-05
Simplifications and improvements to error message rendering.

jswrenn:
This commit simplifies and improves error message rendering.

Rendering is simplified by taking advantage of the cheapness of unstyled `TextMarker`s; CodeMirror display updates are not triggered by `markText` calls that specify neither a `className` or `css` style option. Highlight locations are marked with an unstyled `TextMarker`, highlights visually added and removed by adding a styled `TextMarker` at the same location as the unstyled `TextMarker`.

Pyret source locations appearing in error messages are used to initialize a `Position` object, and the original Pyret `srcloc` value is discarded. Upon initialization, `Position` objects add an unstyled `TextMarker` to the document and their API only discloses the current location of that source range in the document. Liberal use of `Position` objects provides consistent and robust interface for interacting with source locations, and eliminates all bugs caused by using 'stale' source locations.

Test report rendering (`check-ui.js`) has been refactored and is more straightforward.

Highlights may be rendered eagerly (displaying highlights immediately when a message is focused) or lazily (displaying highlights only upon mouseover of the anchor), and with multiple colors or just a single color. A basic user interface allows for mode-switching. Highlighting preference is cached with a session, and highlight display settings can be changed and take effect without re-running the program.

Gutter markers for errored-tests have been temporarily removed. The heuristic used in `master`, `release`, and `horizon` for identifying the point of failure in a check block is not smart enough.


src/web/js/check-ui.js
```diff
@@ -24,246 +24,426 @@
...
+            test.block = this;
+            testFrag.appendChild(test.container);
+          }
+
+          if (error !== undefined) {
+            summary.textContent = "An unexpected error halted the check-block before Pyret was finished with it. " + "Some tests may not have run.";
+            var errorTestsSummary = document.createTextNode("Before the unexpected error, " + tests.executed + (tests.executed === 0 ? " tests " : " test ") + "in this block ran" + (tests.executed > 0 ? " (" + tests.passing + " passed):" : "."));
+            testList.appendChild(errorTestsSummary);
+          } else {
+            summary.textContent = tests.executed == 1 && tests.passing == 1 ? "The test in this block passed."
```
blerner: Please don't use quadruply-nested ternary operators here.  This is almost illegible in JS, even if it's pretty much standard Pyret style to pull the assignment/binding out to the topmost expression ;-)  Use a 4-line if/else if instead.

src/web/js/check-ui.js
```diff
@@ -24,246 +24,426 @@
...
+            //  More than one test; some pass
+            : tests.passing + " out of " + tests.executed + " tests passed in this block.";
+          }
+
+          testList.classList.add("check-block-tests");
+          summary.classList.add("check-block-summary");
+
+          header.classList.add("check-block-header");
+          header.title = "Click to view test results.";
+          header.appendChild(makeNameHandle(name, loc, error !== undefined ? "hsl(0, 100%, 85%)" : tests.executed == tests.passing ? "hsl(88, 50%, 76%)" : "hsl(45, 100%, 85%)").anchor);
```
blerner: Ditto.  Generally speaking, ternary operators don't aid legibility, and long lines like this don't improve their already-tricky clarity

src/web/js/check-ui.js
```diff
@@ -24,246 +24,426 @@
...
+        return;
+        
+      let checkErroredSkeletons = new Array();
+      let testsFailedSkeletons  = new Array();
+      let testsPassedSkeletons  = new Array();
+      
+      let checkResultsContainer = document.createElement("div");
+      checkResultsContainer.classList.add("test-results");
+      try{
+      for(let i = checkBlocks.length - 1; i >= 0; i--) {
```
blerner: Indentation is missing here

src/web/js/check-ui.js
```diff
@@ -24,246 +24,426 @@
...
 
-    // NOTE: MUST BE CALLED WHILE RUNNING ON runtime's STACK
-    function drawCheckResults(container, editors, runtime, checkResults, contextFactory) {
-      var ffi = runtime.ffi;
-      var cases = ffi.cases;
-      var get = runtime.getField;
+        FailingTestSkeleton.prototype.refresh = function refresh() {
+          var snippets = this.tombstone.querySelectorAll(".CodeMirror");
+          for (var i = 0; i < snippets.length; i++) {
+            snippets[i].CodeMirror.refresh();
```
blerner: If there are a large number (i.e. more than ~30) of tests in this skeleton, this loop will cause noticeable jankiness as the browser freezes for a split-second.  Ditto elsewhere.  The example we were using to stress this was a loop `for each(i in range(0, 250)): num-modulo(i, 2) is 1 end`, which easily triggered a lot of failing tests all in one test block (and also all in one line of code).  How can this code avoid the jank?  (Same comment applies elsewhere, for any loop over all snippets' CodeMirrors.)

src/web/js/check-ui.js
```diff
@@ -297,220 +477,90 @@
...
-                                   outputUI.cmPosFromSrcloc(runtime, srcloc, get(test,"loc")))
-                  .wrapper;
-                reason.replaceWith(snippet);
-                block.one("click", ".check-block-header",
-                          function(e) {
-                            snippet[0].cmrefresh();
-                          });
+                skeleton.vivify(
+                  $("<span>").text("An error occurred rendering the reason for this error; details logged to the console"));
+                console.error(result);
```
blerner: Probably worth adding some informative text to this console statement, to help distinguish it from the one on line 512?

src/web/js/check-ui.js
```diff
@@ -297,220 +477,90 @@
...
+
+      container.append($(checkResultsContainer).prepend(summary));
+      
+      }catch(e){console.error(e);}
+      
+      let maybeLocToAST   = outputUI.makeMaybeLocToAST(runtime, documents, srcloc);
+      let srclocAvaliable = outputUI.makeSrclocAvaliable(runtime, documents, srcloc);
+      let maybeStackLoc   = runtime.makeFunction(function(n, userFramesOnly) {return runtime.ffi.makeNone();});
+      
+      function vivifySkeletonFallback(restarter, skeleton) {
```
blerner: Please mark which of these functions must be run on a valid pyret stack, or must be run while not on one.  Stack safety is subtle, and hard to reconstruct the logic for

src/web/js/check-ui.js
```diff
@@ -297,220 +477,90 @@
...
+              runtime.runThunk(
+                function(){
+                  return outputUI.renderErrorDisplay(documents, runtime, result.result, []);
+                }, function(result) {
+                  if (runtime.isSuccessResult(result)) {
+                    if(skeleton.pyretStack)
+                      result.result.append(outputUI.renderStackTrace(runtime, documents, srcloc, skeleton.pyretStack));
+                    skeleton.vivify(result.result);
+                    restarter.resume(runtime.nothing);
+                  } else vivifySkeletonFallback(restarter, skeleton);
```
blerner: Should there be some error message in the console here, that would indicate something went wrong?  This is an utterly silent failure of renderErrorDisplay; likewise the one below is utterly silent about render-fancy-reason...

src/web/js/output-ui.js
```diff
@@ -20,58 +20,153 @@
...
-
-
-    function mapK(inList, f, k, outList) {
-      if (inList.length === 0) { k(outList || []); }
-      else {
-        var newInList = inList.slice(1, inList.length);
-        f(inList[0], function(v) {
-          mapK(newInList, f, k, (outList || []).concat([v]))
+    
+      function Position(doc, source, from, to, inclusiveLeft, inclusiveRight) {
```
blerner: Is this duplicate code of the other Position function plus the rangeHandle code in beforePyret.js?

src/web/js/output-ui.js
```diff
@@ -134,278 +218,55 @@
...
-      // be either fadeAmt or zero.
-      var warnDesired = 0;
-      // These are just fixed parameters and can be eliminated if
-      // efficiency demands it.
-      var warnWait = 250;
-      var warnDuration = 5000;
-      var fadeAmt = 0.5;
+        var direction;
+        var TOP     = 0,
+            BOTTOM  = 1;
```
blerner: @jpolitz an overall style question: how do we want to handle multiple definitions per `var` line?  emacs has a bear of a time indenting and highlighting them correctly...

src/web/js/output-ui.js
```diff
@@ -612,70 +394,46 @@
...
-    function makeMaybeLocToAST(runtime, editors, srcloc) {
+    function makeMaybeLocToAST(runtime, documents, srcloc) {
       return runtime.makeFunction(function(loc) {
         return runtime.ffi.cases(runtime.getField(srcloc, "is-Srcloc"), "Srcloc", loc, {
           "builtin": function(_) {
             console.error("maybeLocToAST should not be passed a builtin source location.", loc);
             return runtime.ffi.makeNone();
           },
-          "srcloc": function(filename, start_line, start_col, startCh, endL, endC, endCh) {
+          "srcloc": function(filename, start_line, start_col, _, end_line, end_col, _) {
```
blerner: I think it's illegal to have a JS function with two arguments of the same name, even if that name is "I don't care!" -- possibly worth restoring these two param names

src/web/js/output-ui.js
```diff
@@ -685,278 +443,234 @@
...
-          if(editor) {
-            styles.insertRule(
-                  ((context === undefined) ? "" : "#main[data-highlights=" + context + "]")
-                  + " ." + locKey + " { background-color:" + cssColor + ";}",
-                  0);
-            for(var i=cmLoc.start.line;i<=cmLoc.end.line;i++){
-              editor.addLineClass(i,"background",locKey);
-            }
-            return locKey;
+    var Snippet = function () {
```
blerner: Why the extra thunking here?

test-util/util.js
```diff
@@ -216,10 +216,10 @@ function testRunsAndHasCheckBlocks(it, name, toEval, specs) {
       self.browser.wait(function () {
         return self.browser.isElementPresent(webdriver.By.className("check-results-done-rendering"));
       }, 20000);
-      return response.findElements(webdriver.By.className("check-block-result"));
+      return response.findElements(webdriver.By.className("check-block"));
     });
     var blocksAsSpec = checkBlocks.then(function(cbs) {
-      var tests = cbs.map(function(cb, i) {
+      var tests = cbs.slice(1).map(function(cb, i) {
```
blerner: What's this for?

blerner:
On first look, the code seems reasonable.  I haven't built or run it yet, so I don't know exactly what it does.  I left mostly stylistic comments so far; I'd like to read the code more carefully a second time.  

I'm glad to see that there's no longer a use of insertRule; that immediately clears up one problem we'd talked about.  I'm still a bit concerned about too many CM instances and looping over them freezing the browser, though; I'd like to hear more about that design.

Also, do you know why the travis builds failed above?

jswrenn:
Still on break, apologize for terse response. Selenium tests pass locally;
no idea why Travis is timing out.

I saw some CM benchmarks and it seemed to handle creating lots of instances
just fine.

Most are not visible when created so the display update doesn't happen
until the check block is expanded.

src/web/js/check-ui.js
jswrenn: @blerner: Unlike previous encounters with jank, the jank here is only encountered when the block is expanded; you don't pay for displaying tests you can't see.

The thing to do is a bit of benchmarking. If the issue is that the browser is reflowing/repainting over-eagerly, a good solution would be to push the snippets that need refreshing into a work queue and use 
I think a good solution would be to push the snippets that need refreshing into a work list and then use a bunch of calls of `requestAnimationFrame` to churn through them.

src/web/js/output-ui.js
jswrenn: It's allowed and well addressed in the specification, though this particular use is disallowed on strict-mode functions. I removed the names after a particularly frustrating bug caused by a naming inconsistency: CodeMirror positions refer to the the text column component of the position as `ch`, but in Pyret `srcloc` values, the `ch` fields are indices into the characters of the entire document.

(I think I recall the `ch` fields of pyret srclocs actually being incorrect, but I can't recall how I encountered that... Regardless, they're of absolutely no use.)

I could rename the underscores `_start_ch` and `_end_ch` following the prolog convention of `_name` indicating an unused and ignored parameter, but in javascript the convention  of prefixing names with an underscore usually means something entirely different.

src/web/js/output-ui.js
jswrenn: @blerner In the days when variable hoisting was unavoidable*, thunks were used in javascript to allow local variable definitions without polluting the outer namespace. Old habits die hard, and I followed this pattern for class declarations. It's also how babel.js desugars `class` declarations.

It's not strictly necessary, but it makes it visually distinct from normal function declarations and provides a good foundation for expanding the interface (a `refresh` method would be a good edition, for example, to shorten the `someSnippet.editor.refresh()` lookup and calls).

*I accidentally let a few `let`s slip through. Are they kosher to use, yet?

test-util/util.js
jswrenn: I moved the summary to the top of the repl (since results are added eagerly, from top-to-bottom, roughly) and its class name is `check-block`.

jpolitz:
One thing to work on right away is the interface for choosing the highlighting mode.  No way can it go where it is; it won't fit once logged in on smaller monitors, and there's no reason highlighting mode should get the most prominent real estate in the interface.

I think putting it under `More...` is the best bet.

src/web/js/beforePyret.js
```diff
@@ -48,15 +48,127 @@ window.stickMessage = function(message) {
...
+    inclusiveRight = false;
+    shared = true;
+  } else {
+    if (options.inclusiveLeft !== undefined)
+      inclusiveLeft = options.inclusiveLeft;
+    if (options.inclusiveRight !== undefined)
+      inclusiveRight = options.inclusiveRight;
+    if (options.shared !== undefined)
+      shared = options.shared;
+  }
```
jpolitz: Is it intentional that `options` _must_ have all three fields if it is provided?  If only one or two can be provided, this flow seems wrong, because if it's, say, `{inclusiveLeft: true}`, `inclusiveRight` and `shared` will be left `undefined`.

src/web/js/beforePyret.js
```diff
@@ -48,15 +48,127 @@ window.stickMessage = function(message) {
...
+  return this.markText(position.from, to,
+    { inclusiveLeft   : inclusiveLeft,
+      inclusiveRight  : inclusiveRight,
+      shared          : shared  });
+});
+
```
jpolitz: Lines 80-86 set `Position` to `undefined`, right?  Unless there's something special about this context that I'm not getting; I get `undefined` for `Position` if I use this at the console.

Is this dead code, and some other definition of `Position` is actually being used?  There seem to be a bunch of things named `Position`.

src/web/js/beforePyret.js
```diff
@@ -48,15 +48,127 @@ window.stickMessage = function(message) {
...
+var Position = function() {
+  function Position(doc, from, to) {
+    this.doc  = doc;
+    this.changeGeneration = doc.changeGeneration();
+    this.from = from;
+    this.to   = to;
+  }
+}();
+
+var Documents = function() {
```
jpolitz: I don't particularly like this pattern of `var Thing` and then an IIFE with a `function Thing` inside it.  The shadowing is pretty odd to think about, IMO.

src/web/js/check-ui.js
```diff
@@ -24,246 +24,426 @@
...
+          header.appendChild(makeNameHandle(name, loc, error !== undefined ? "hsl(0, 100%, 85%)" : tests.executed == tests.passing ? "hsl(88, 50%, 76%)" : "hsl(45, 100%, 85%)").anchor);
+
+          container.classList.add("check-block");
+          container.classList.add(error !== undefined ? "check-block-errored" : tests.executed == tests.passing ? "check-block-success" : "check-block-failed");
+          container.appendChild(header);
+          testList.appendChild(testFrag);
+          container.appendChild(summary);
+          container.appendChild(testList);
+
+          var tombstone = undefined;
```
jpolitz: I don't grok the name `tombstone`, but it seems important.  Is it because it indicates an error, or some other reason?

src/web/js/error-ui.js
```diff
@@ -55,6 +55,46 @@
       }
 
       // HELPERS
+      // NOTE: MUST BE CALLED WHEN RUNNING ON runtime's STACK
+      function renderSimpleReason(err) {
+        var ret = Q.defer();
+        runtime.runThunk(function() { 
+          return get(err.exn || err, "render-reason").app();
```
jpolitz: Oof.  This `err.exn || err` is a real annoyance.  I know it's because there are inconsistent representations coming from somewhere else.  I wonder if a single helper that takes an "error-shaped" thing and produces a Pyret object for it would be useful.  I'm worried about these short little hacks littering the code and leading to accidental undefined values.

src/web/js/error-ui.js
```diff
@@ -137,7 +176,7 @@
               .append($("<p>").text("Error: "))
               .append(exnstringContainer)
               .append($("<p>"))
-              .append(outputUI.renderStackTrace(runtime, editors, srcloc, e));
+              .append(outputUI.renderStackTrace(runtime, documents, srcloc, e.pyretStack));
```
jpolitz: Was this just a bug before this diff, or did the signature of `renderStackTrace` change?

src/web/js/error-ui.js
```diff
@@ -151,107 +190,59 @@
...
                       .text("An error occurred rendering the reason for the above error; details logged to the console");
                     if(!container.hasClass("internal-error")) {
-                      drawError(errorError, editors, runtime, errorDisp, undefined);
+                      drawError(errorError, documents, runtime, errorDisp, undefined);
                     }
                     container.append(errorError);
-                    errorError.trigger('toggleHighlight');
+                    errorError.click();
                     console.error("drawPyretRuntimeError: render-fancy-reason failed:", errorDisp);
                     console.log(e.exn);
```
jpolitz: Another bare log that could use some explanation.

src/web/js/output-ui.js
```diff
@@ -20,58 +20,153 @@
...
-    function mapK(inList, f, k, outList) {
-      if (inList.length === 0) { k(outList || []); }
-      else {
-        var newInList = inList.slice(1, inList.length);
-        f(inList[0], function(v) {
-          mapK(newInList, f, k, (outList || []).concat([v]))
+    
+      function Position(doc, source, from, to, inclusiveLeft, inclusiveRight) {
+        var inclusiveLeft   = false,
+            inclusiveRight  = false;
```
jpolitz: These `var` declarations shadow the function parameters, so lines 46 and 48 can never have their conditions be true.  So `this.inclusiveRight` and `this.inclusiveLeft` are always `false`, I believe.

jpolitz:
General style comment:  I've consistently tried to encourage folks to _always_ use braces around the body of `if` statements, and avoid one-liners like this:

```
if (condition)
  statement
```

It's far too easy to add another statement there and forget that it isn't guarded by the conditional because the braces were forgotten.  It would be great to have this consistent across the codebase.

src/web/js/output-ui.js
```diff
@@ -20,58 +20,153 @@
...
+        });
+        
+        Object.defineProperty(this, 'to', {
+          get: function() {
+            var pos = cached_find(doc, positionCache, textMarker);
+            return pos !== undefined ? pos.to : undefined;
+          }
         });
+        
+        positionCache.set(doc.changeGeneration(), {from, to});
```
jpolitz: It might be good to skip ES6 features for now; I'm not super-confident in all our transpiling for `/editor`.

src/web/js/output-ui.js
```diff
@@ -134,278 +218,55 @@
...
+      $(".warning-upper.hinting, .warning-lower.hinting").removeClass("hinting");
+      
+      var editor = position.doc.getEditor();
+      
+      if (!(editor instanceof CodeMirror))
+        throw new Error("Source location not in editor", position);
+      
+      var coord = editor.charCoords(
+        {line: position.from.line, ch:0},
+        position.source === "definitions://" ? "local" : "page");
```
jpolitz: It would be nice to make these ad hoc enums be constants defined elsewhere, as e.g. `LOCALSOURCE = "local", PAGESOURCE = "page"`, since presumably they are duplicated string literals.

src/web/js/output-ui.js
jpolitz: I usually do one `var` statement per, because it makes moving code around later easier.  But I don't have strong feelings about it.

src/web/js/output-ui.js
```diff
@@ -550,35 +335,34 @@
...
-    function emphasizeLine(editors, cmloc) {
-      var editor = editors[cmloc.source];
-      if(!editor) return;
-      for(var i = cmloc.start.line; i <= cmloc.end.line; i++) {
-        setTimeout(
-          function(){editor.removeLineClass(this, "background", "emphasize-line");}
-            .bind(editor.addLineClass(i, "background", "emphasize-line")),
-          500);
+    
+    function drawPosition(position) {
```
jpolitz: Can `drawPosition` and `drawSrcloc` be harmonized/have their common parts factored out?  They seem very similar.

src/web/js/output-ui.js
jpolitz: We need to keep all of our code strict-mode compliant.

When I want to do this, I just use increasing numbers of underscores, so `_` for the first, `__` for the second, and so on.

src/web/js/output-ui.js
```diff
@@ -592,17 +376,15 @@
     var goldenAngle = 2.39996322972865332;
     var lastHue = 0;
     
-    function makeSrclocAvaliable(runtime, editors, srcloc) {
+    function makeSrclocAvaliable(runtime, documents, srcloc) {
```
jpolitz: This must be called in Pyret stack contexts, right?  Due to the use of `pauseStack` below.  Making a note when I see functions like this.

src/web/js/output-ui.js
```diff
@@ -612,70 +394,46 @@
...
-                    restarter.resume(runtime.ffi.makeNone());
-                  }
-                });
+            var source = documents.
+              get(filename).
+              getRange(
+                new CodeMirror.Pos(start_line - 1, start_col),
+                new CodeMirror.Pos(  end_line - 1,   end_col));
+            runtime.pauseStack(function(restarter) {
+              runtime.run(function(_, __) {
```
jpolitz: This ought to be `runThunk`, since none of the parameters are being used, and the `sync: false` is redundant in CPO, which forces async execution.

src/web/js/output-ui.js
```diff
@@ -612,70 +394,46 @@
...
+                return runtime.getField(PP, "surface-parse").app(prelude + source, filename);
+              }, runtime.namespace, {
+                sync: false
+              }, function(result) {
+                if(runtime.isSuccessResult(result)) {
+                  restarter.resume(runtime.ffi.makeSome(result.result.dict.block.dict.stmts.dict.first));
```
jpolitz: This chain of accesses almost certainly needs to be a well-named function, like `getFirstExpressionFromProgram`.

jpolitz: Also, this fails badly on the empty program.  I know that the empty program "can't possibly have any errors," except I'm sure there's some way to make it happen by messing up default includes, so we should make sure that doesn't fudge up.

src/web/js/output-ui.js
jpolitz: Yeah, I dislike the pattern, but I'm not going to argue strongly against if it's used consistently.

jswrenn:
@jpolitz:
> it won't fit once logged in on smaller monitors

At the time of pull request, I believe it displayed correctly on 1280x1024 displays, which were common in my high school. I've reduced the font size slightly, and it now displays without wrapping on 1024x768 displays:

XGA display

Displays smaller than 1024x768 constitute only about [3% of the visitors to W3Schools](http://www.w3schools.com/browsers/browsers_display.asp). Can you take a look at what Google Analytics has to say about our visitor's browser resolution? What's the lower bound of resolutions we'd like to support?

> there's no reason highlighting mode should get the most prominent real estate in the interface. I think putting it under More... is the best bet.

The space in the center of the toolbar might be "prime", but it's also otherwise unused. For the sake of comparing the effectiveness of these different options, it is absolutely essential that they be exceedingly discoverable, comprehensible, and inviting to play with. The top-bar location ticks all of these boxes.

The `More...` menu is well suited to accommodating a pair of check boxes, but not a sentence. I tried to express these options like that at first. There isn't room in the menu for descriptive labels, so the addition of help icons that can display more information is all but necessary.  The single-sentence approach is immediately comprehensible and approachable. The unused space in the toolbar accommodates this form-factor perfectly, and makes it exceedingly discoverable.

shriram:
Are you going to have a separate config for people on tablets?

jpolitz:
That screenshot isn't logged in (the scenario I described), where the program name and save and open buttons make the toolbar much wider.

Also, on a projector, the resolution is very often less than 1024x768, and Pyret on a projector is an extremely common use case.

Folks also zoom in one or two ticks, or make the window half or two-thirds the size of their monitor.  We already deal with the Run button popping onto the next line with some of these use cases.

Separately, I disagree that this is a discoverable interface.  As a new user, I wouldn't know what highlights mean, or why there should be colors picked.  If my program is fine, then these choices evidently do _nothing_, which might lead me to simply ignore them in the future.

A discoverable interface would pop up the first time the user has a highlighted error, and would guide them through the different options, and then show them how to choose and how to find the options in the future.  In fact, having the options in the location that "Highlight This..." used to be might accomplish this well, since it puts the options near the thing they affect.

jswrenn:
@shriram: I [checked common tablet sizes](http://mydevice.io/devices/#sortTablets) by the numbers, this works without wrapping on all of them in landscape mode, and on some tablets in portrait mode. 

---

> Folks also zoom in one or two ticks, or make the window half or two-thirds the size of their monitor. We already deal with the Run button popping onto the next line with some of these use cases.
> Separately, I disagree that this is a discoverable interface.

@jpolitz There are two senses in which something is discoverable, and the first is being aware that such an option exists. I keep my CPO window zoomed in roughly two ticks at all times because Ctrl+Plus is a shorter path to making my source code legible than finding the font options in the 'More' menu (which I often forget exists at all). The top-bar placement is conspicuous and accessible. 

---

> A discoverable interface would pop up the first time the user has a highlighted error, and would guide them through the different options, and then show them how to choose and how to find the options in the future.

That sounds like a great mechanism, but with a conspicuous and almost self-explanatory interface it can be of a much lighter weight: just a tool-tip below the option saying
``` ⬑ Change how we show you errors. ```

---

> Also, on a projector, the resolution is very often less than 1024x768, and Pyret on a projector is an extremely common use case.

Would 800x600 be a reasonable target resolution to support? Without knowing the minimal resolution we'd like to support or the resolutions of our users, I can't intentionally design an interface that accommodates them.

File: src/web/js/check-ui.js (hunk 2 above)
Comment (jswrenn) [inline src/web/js/check-ui.js]: A lot of these quirks address code that was introduced pre-pull request but called out in the diffs because so much was re-intended, moved around, etc. Since this pull request comes after the merge to master, a goal was to change as little as possible and to limit the changes I did introduce to things that overall simplified the code and reduced the surface-area for bugs.

File: src/web/js/output-ui.js (hunk 10 above)
Comment (jswrenn) [inline src/web/js/output-ui.js]: > We need to keep all of our code strict-mode compliant.

Should we be adding `"use strict";` to our javascript files, then?

File: src/web/js/check-ui.js (hunk 16 above)
Comment (jswrenn) [inline src/web/js/check-ui.js]: `tombstone`s are elements that are part of the initially rendered test results skeleton, and are later inscribed in with a postmortem.

It just happened to be the word that came to mind when I was working with skeletons. ;)

File: src/web/js/error-ui.js (hunk 17 above)
Comment (jswrenn) [inline src/web/js/error-ui.js]: > I'm worried about these short little hacks littering the code and leading to accidental undefined values.
Having a handful of shapes that errors can come is the root of the issue here (and the result of most of the verbosity of `error-ui.js`; some of the structures that file accommodate before the error overhaul weren't used anywhere, even then). A consistent shape for error values would simplify things in a more maintainable way.

File: src/web/js/error-ui.js (hunk 18 above)
Comment (jswrenn) [inline src/web/js/error-ui.js]: The signature changed to accept the stack array, rather than the error value.

**Comment by jpolitz** [comment]
I don't have a spec for the width that things have to be, or user data describing a min/max range.  It would be good to have or generate such specifications and data, but they don't exist at the moment, and the lack of it shouldn't cause us to throw up our hands and make unreasonable decisions.  Given that I personally have probably looked at the most different active monitors that have CPO open, and I've observed that horizontal space is at a premium, I humbly ask that you consider my opinion at least somewhat relevant for this decision, in spite of the lack of hard data.

800x600 may be sufficient, but _any_ use of horizontal space comes with a cost, given that we already have issues with it.  It's simply a challenge with the current interface, and is exacerbated by anything that adds fixed horizontal width.


The implicit argument here is that studies deserve dedicated UI space, which I'm sympathetic to, though it's a brand-new idea, so there are a lot of issues to consider that we haven't discussed before.

-  How long is this interface intended to last?  It seems shortsighted to have this interface take up this part of the screen forever, because by this argument, if CPO is ever used to study something else, the prominent space ought to be used for that study.  In addition, I find it highly unlikely that error highlighting will for all time be the thing that ought to be most prominent in the interface, so it ought to move some time after the needs of the study are met.
- Will the results of the study still hold with the interface presented significantly differently in the future?
- I hypothesize that with this interface, some amount of training will still be required for students to understand what the options mean.  Do you agree with that?  Will there be a presentation, web site, or other kind of instruction on using the interface?
- If it's put up, we're sure to need an announcement describing it on -discuss, and I would anticipate lots of questions about what it's for and what it does from folks who don't see that announcement, but see it later.  Their first impression of Pyret will be that the thing we care most about is options for error display, which is almost certainly confusing and irrelevant to them at first.  I'm not sure whether that's good or bad, but it's certainly a significant impact, and far from a slam-dunk marketing and usability decision I'd choose in a vacuum.
- If this space is to be used for limited-duration events, affects all Pyret users, and requires special explanation, we need a plan for _presenting it_ and _framing it_.  What does that plan look like?

File: src/web/js/output-ui.js [hunk 27]
```diff
@@ -685,278 +443,234 @@
...
+    function highlight(color) {
+      if(colorsHighlighted.has(color))
+        return;
+      else {
+        colorsHighlighted.add(color);
+        var anchors   = allHighlightAnchors.get(color);
+        var positions = allHighlightPositions.get(color);
+        for(var i = 0; i < anchors.length; i++) {
+          anchors[i].css('background-color',
+            sessionStorage.getItem("highlight-colorfulness") != "vibrant"
```
jpolitz: I think directly fetching from `sessionStorage` here is a little suspect.  First, it's in a loop, so the call happens quite a bit more often than necessary.  Second, it makes it harder to parameterize highlighting in different ways than fetching directly from storage.  Would it work to have `colorfulness` be a parameter to `highlight` instead?

src/web/js/output-ui.js
```diff
@@ -1159,92 +877,63 @@
...
-                }
+              if(!messageAnchors.has(color))
+                messageAnchors.set(color, [anchor]);
+              else messageAnchors.get(color).push(anchor);
+              if(!messagePositions.has(color))
+                messagePositions.set(color, positions);
+              else Array.prototype.push.apply(messagePositions.get(color), 
+                                              positions);
+              anchor.on("click", function (e) {
+                if (positions[0] !== undefined)
```
jpolitz: Is this check catching the case where `positions` could be _empty_?   It doesn't seem like `fromPyretSrcloc` can return `undefined`, so should this be a check for `length`?  Or should the `"click"` event simply not be registered if `positions` is empty?

src/web/js/output-ui.js
```diff
@@ -1159,92 +877,63 @@
...
+                event.stopPropagation();
               });
-              
-              anchor.on("mouseleave", function() {
-                for (var i = 0; i < locClasses.length; i++) {
-                  unhintLoc(runtime, editors, srcloc, locArray[i]);
-                  $("."+locClasses[i]).css("animation", "");
-                }
+              anchor.on("mouseenter", function () {
+                if (positions[0] !== undefined)
```
jpolitz: Ditto the comment above about positions being empty.

src/web/js/output-ui.js
```diff
@@ -1256,57 +945,38 @@
...
+
+        rendering.bind('toggleHighlight',function() {
+          colorsHighlighted.forEach(function(color) {
+            unhighlight(color);
           });
-        }
+          colorsEmphasized.forEach(function(color) {
+            demphasize(color);
+          });
+          if(sessionStorage.getItem('highlight-eagerness') != 'lazy') {
```
jpolitz: Ditto another comment about using `sessionStorage` in the middle of other code – probably better to parameterize over this.

src/web/js/output-ui.js
```diff
@@ -1411,6 +1081,7 @@
           });
           return container;
         } else {
+          return runtime.getField(O, "some").app(v);
```
jpolitz: Dead code?

src/web/js/repl-ui.js
```diff
@@ -319,13 +320,13 @@
...
 
-        editors = {};
-        editors["definitions://"] = uiOptions.cm;
-        function invalidateHighlights(cm, change) {
-          cm.off("change", invalidateHighlights);
-          document.getElementById("main").dataset.highlights = "";
-        }
-        editors["definitions://"].on("change", invalidateHighlights);
+        CPO.documents.forEach(function(doc, name) {
+          if (name.indexOf("interactions://") === 0)
```
jpolitz: General comment that it would be good to define a constant for `"interactions://"` and `"definitions://"` to use in more places.  I'm guilty of this, too, and it made some changes in early summer harder to make.

test-util/util.js
jpolitz: For posterity, the slice is removing the div containing "5 passed, 3 failed" in:

```
|-------------------------|
  5 passed 3 failed
|-------------------------|

|-check-block-1-----------|
  line 54 passed
|-------------------------|

|-check-block-2-----------|
  line 56 passed
|-------------------------|
```

src/web/css/editor.css
```diff
@@ -625,6 +625,72 @@ div#font-label:hover{
   box-shadow:none;
 }
 
+.highlight-settings {
```
blerner: A general CSS issue we noticed: there's a rule that gives all `error ul` elements a non-zero padding.  That rule unfortunately supercedes the rule on `ul.inlineCollection` that specifies zero-width padding.  The upshot is that `[list:   empty]` renders with too much whitespace in it.

jpolitz:
When beginning editing with more than a few dozen errors/check results rendered at once, there's a noticeable lag on the event that clears and sets the "stale" message.  I think adding a `CM.operation` may be able to help here.

jpolitz:
On editing, the underlines under check results/errors and the gutter highlights don't seem to clear.

jpolitz:
I don't see any difference between the behavior when the different options are selected.  Choosing one color, for instance, and re-running this program, just shows the multi-color highlighting:

screen shot 2016-08-04 at 8 36 01 am

jpolitz:
Deleting my last few comments because I'm not sure I was in a clean enough build state.

jpolitz:
Overlapping highlights are a little weird; the snippet referring to the whole expression here didn't get a color, and there's some jank when highlighting the "left operand":

screen shot 2016-08-16 at 6 08 43 pm

src/web/js/check-ui.js
jswrenn: Addressed by be1bf70a4541e90da8e67b5d050728a46f3bf45a.

src/web/js/check-ui.js
jswrenn: Kind-of addressed by 74acb3e311eee465bcc8b3381943c0fcbef64963

src/web/js/output-ui.js
jswrenn: Yes. Addressed by 887c8cbd38df600075446f45a6eee0f6bc94deb9

src/web/js/output-ui.js
jswrenn: @jpolitz Good catch! `Position` used to take an option object with fields of those names, but when I added them as parameters, I did not rename the local variables. (This doesn't cause any bugs, since `Position` is always instantiated with exclusive bounds at the moment, but that will change.)

src/web/js/output-ui.js
jswrenn: @jpolitz Addressed by 60517497cca81e2413fc236eeb3160a0412e00f9

This was purely a mistake. TIL about object shorthand.

jswrenn:
@jpolitz CodeMirror doesn't behave well with simultaneous nested highlights, so I try to avoid them wherever possible. One work-around is using snippets, another is using hinted highlights: regardless of highlighting mode, the highlight corresponding to "field lookup expression" is only shown on mouse-over.

The "left operand" anchor is also a hinted highlight, because if we rendered its highlight concurrently with the other two highlights in that expression, we'd have `{}` in one color, `.` in another color, and `f` in another color. The highlight on `.` is actually the one that encompases the other two highlights, but instead it gets visually sandwiched.

src/web/js/check-ui.js
jswrenn: 3998c362a4900f862425df3dfa6e6f5384c20681 addresses this using `requestAnimationFrame` (but not quite as described above). The pause in expanding the check block is basically eliminated.

src/web/js/output-ui.js
jswrenn: @jpolitz Addressed by 4f699ab746a209288ab1def6bca87ad5d2b1c640. Some of the changes get a little silly, such as:
```
function(filename, _, __, ___, ____, _____, ______) { 
  ...
```

It would be valid javascript to take advantage of loosey-goosy arity and write
```
function(filename) { 
  ...
```
but that seems like playing with fire.

src/web/js/output-ui.js
jswrenn: Addressed by 71b670dd1ad76e0e5de3a7ae988fb19ef8747ad5

src/web/js/output-ui.js
jswrenn: a0fdb1b3a09c88a9c4fa4b6a994277a2af1b5ce3 addresses the poor failure behavior.

src/web/js/output-ui.js
jswrenn: Partially addressed by 1d5fc0fe85e954303acd5ca56735208373c5bf0d. This lifts the `sessionStorage` access out of the loop, but not parameterize the functions. I'm a little wary of working to support a feature that doesn't have an immediate or foreseeable need. Parameterizing the functions is straightforward, but manipulating highlight mode with some other mechanism would require a good deal of careful thought. It seems prudent to cross that bridge when we get there.

src/web/js/check-ui.js
```diff
@@ -189,8 +189,9 @@
 
         FailingTestSkeleton.prototype.refresh = function refresh() {
           var snippets = this.tombstone.querySelectorAll(".CodeMirror");
-          for (var i = 0; i < snippets.length; i++) {
-            snippets[i].CodeMirror.refresh();
+          for (let i = 0; i < snippets.length; i++) {
+            window.requestAnimationFrame(
```
blerner: Would it be better to hoist the request out of the loop, and refresh all the cm instances at once?

src/web/js/check-ui.js
```diff
@@ -504,14 +506,14 @@
               } else {
                 skeleton.vivify(
                   $("<span>").text("Two errors occurred rendering the reason for this error; details logged to the console"));
-                console.error(result);
+                console.error("Displaying rendered reason failed in `vivifySkeletonFallback`:", result);
```
blerner: :+1:

src/web/js/output-ui.js
```diff
@@ -401,25 +401,30 @@
...
-              runtime.run(function(_, __) {
+              runtime.runThunk(function() {
                 return runtime.getField(PP, "surface-parse").app(prelude + source, filename);
-              }, runtime.namespace, {
-                sync: false
               }, function(result) {
                 if(runtime.isSuccessResult(result)) {
-                  restarter.resume(runtime.ffi.makeSome(result.result.dict.block.dict.stmts.dict.first));
+                  try {
+                    restarter.resume(runtime.ffi.makeSome(result.result.dict.block.dict.stmts.dict.first));
```
blerner: Technically, not quite right: this will also catch failures in the subsequent resumption and claim they were the fault of `makeSome(result....first)`.  Better to use a local variable here just around the field extraction, or better yet, just use a chained 
```
var res = result;
res = res && res.result;
res = res && getField(res, "block");
res = res && getField(res, "stmts");
res = res && getField(res, "first");
if (res) { restarter.resume(makeSome(res)); }
else { console.error ... }
```

src/web/js/output-ui.js
```diff
@@ -542,17 +547,13 @@
         colorsHighlighted.add(color);
         var anchors   = allHighlightAnchors.get(color);
         var positions = allHighlightPositions.get(color);
+        var colorfulness = sessionStorage.getItem("highlight-colorfulness");
         for(var i = 0; i < anchors.length; i++) {
           anchors[i].css('background-color',
-            sessionStorage.getItem("highlight-colorfulness") != "vibrant" 
-                ? globalColor
-                : color);
+            colorfulness != "vibrant" ? globalColor : color);
```
blerner: why not pull this conditional out, too?

src/web/js/output-ui.js
jswrenn: @blerner Doesn't `getField` throw an exception if the field is not found?

src/web/js/check-ui.js
jswrenn: > Would it be better to hoist the request out of the loop, and refresh all the cm instances at once?

Jank is what happens when the browser is too caught up running javascript to repaint effectively. If my understanding of `requestAnimationFrame` is correct, using multiple calls of `requestAnimationFrame` leaves the decision of how refreshes and repaints are interleaved up to the browser. If `refresh` is very quick, the browser may be able to churn through several `requestAnimationFrame` before it paints the frame. Using a monolithic call of `requestAnimationFrame` runs the risk of having an extremely long frame.

On the other hand, since these are only snippets being refreshed here, it's probably safe to assume that the `refresh` calls are very quick. I haven't profiled enough to say for sure.

src/web/js/output-ui.js
blerner: ok, then don't use getField; say res.dict.block instead.

src/web/js/output-ui.js
jswrenn: Addressed by 49565f34a66fae3b34b09c83be1562bad2242f51

src/web/js/output-ui.js
jswrenn: Addressed by 78ca1e80e8cc9ef0cbd2c5f46226cc222aa0fb2f

jpolitz:
Whoa all the logging stuff is part of this now?  That makes it much harder to stage.

jswrenn:
> Whoa all the logging stuff is part of this now? That makes it much harder to stage.

@jpolitz Yeah, it's awkward. The error-ui.js changes made in #122 was necessary for changing the mode switcher interface so that every message had a settings icon. So I wrapped up the logging implementation, merged `logging` into `errors-horizon` and then made the mode switcher changes (598c0d8a2c19a81332df7b79d4ec44a351de1127).

jswrenn:
be43926b9b60980fc175accd736b9ba9512b3599 adds a dummy backend (`logger.log` does nothing), so that we can think about integrating this PR without blocking on having a storage service in place.

PR #106, opened 2016-08-13
Enable support for code-folding in CPO editor

belph:
This Pull Request adds the ability to fold regions of code inside of the CPO editor. *With the exception of parentheses* (since that just clutters the gutter), this adds the ability to fold code between any pair of matched keywords/delimiters that CPO currently highlights.

Examples:

No folding:

![Code unfolded][fold-1]

Folding of inner `for...end` expression (note the toggling button on the side gutter):

![for...end folded][fold-2]

As a special case, `fun` and `method` keep their names visible:

![fun folded][fold-3]

Possible enhancements:
- Make `data` keep its name visible (requires not-exactly-trivial finagling of `parse` in `pyret-mode.js`)
- Keep arguments visible in `fun`/`method` (perhaps just keep everything between keywords and their matching colons visible?)
- Add keybinding to toggle folding

[fold-1]: https://belph.github.io/photo/fold1.png
[fold-2]: https://belph.github.io/photo/fold2.png
[fold-3]: https://belph.github.io/photo/fold3.png

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,10 @@ build/web/neweditor/js/%.js: src/web/neweditor/js/%.js
 build/web/css/codemirror.css: $(CM)/lib/codemirror.css
 	cp $< $@
 
-MISC_CSS = build/web/css/codemirror.css
+build/web/css/foldgutter.css: $(CM)/addon/fold/foldgutter.css
+	cp $< $@
+
+MISC_CSS = build/web/css/codemirror.css build/web/css/foldgutter.css
 
 COPY_GIF := $(patsubst src/web/img/%.gif,build/web/img/%.gif,$(wildcard src/web/img/*.gif))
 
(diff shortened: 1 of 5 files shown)


belph:
Also worth noting: in ambiguous cases, the *outermost, leftmost* block is chosen. 
Consider the following code snippet:
```
fun foo(x): for map(elt from x): elt + 1 end end
```
This would be folded as
```
fun foo<->end
```
Additionally, the following code
```
fun bar(x): x + 1 end fun baz(x): x + 2 end
```
would be folded as
```
fun bar<->end fun baz(x): x + 2 end
```

...but if your code looks like that, no amount of folding will help you :)

shriram:
Thank you!!!

blerner:
To handle `data name<->end` folding, tweak around line 200, the "Level 2" part of `tokenBase`.  We have enough information to tell whether the current token is a name and the previous token is `data`.  If so, then change the style from `"type"` to `"datatype"`, and then you can examine the `style` of tokens elsewhere in your code.  (You'll also have to tweak editor.css to include `cm-datatype` as a synonym of `cm-type`, because that's where the styles eventually get used for rendering, but that's easy.)

Issue #108, opened 2016-08-15
Parameter to control the radius of a pie chart


jpolitz:
Right now, pie charts decide on their size when created; it would be useful to control their size with a radius parameter.

blerner:
In the group discussion last week, they also requested being able to resize the chart in the popup window by resizing the popup window.

jpolitz:
Do you have a sense of "would be nice" vs. "can't do my course without" for the various requests?

blerner:
This one was a medium-strong request: these teachers use pie charts to indicate the energy budget of a system, with the relative wedge sizes indicating e.g. the proportions of kinetic to potential to chemical to heat energy, say.  But then they use the radius to indicate the actual total amount of energy, with "bigger radii == more energy".  So they really wanted some visceral action to "add energy to the system".

sorawee:
OK. In Google Charts, there's nothing like this. The closest thing I can get is the exploded pie chart which looks like this.

![screenshot from 2017-07-23 16-15-30](https://user-images.githubusercontent.com/9099577/28502525-2ccd2502-6fc2-11e7-82ec-2fd9150a763d.png)

Is this acceptable?

blerner:
I don't think they wanted individual radii per slice; I thought they wanted an overall radius set for the whole thing.  But I'm really not sure anymore.

sorawee:
Since resizing will make the chart bigger/smaller, is it enough to say please resize it?

Issue #111, opened 2016-08-17
Bad line-plot


sorawee:
```
include plot

display-line('test', table: x, y
    row: 100, 100
    row: 100, 101
  end)
```

results in 

```
An unexpected error occurred: TypeError: Cannot read property 'length' of undefined
```

blerner:
Well, this is better(?) now: the same program results in 
```
Evaluating a expression in <builtin builtin://plot> errored.

It was expected to produce a "Pyret Value", but it produced a non-"Pyret Value" value:
    nothing
(Hide program execution trace...)

Evaluation in progress when the error occurred (most recent first):
builtin://plot:508:2-508:17
builtin://plot:397:2-399:21
definitions://:2:0-5:6
    display-line('test', table: x, y
```
Still seems wrong, though -- `nothing` is indeed a Pyret value, so something else is haywire here  @sorawee is this code supposed to work? And if it *isn't*, then what should the intended error message be?  And why is it getting `nothing` instead?

sorawee:
@blerner I need to take a look a little bit more, but it seems that the value returned from `plot-lib.js` is `undefined`. For example, try:

```
import plot-lib as P

P.histogram({x-axis: "", y-axis: "", xscale: 0, yscale: 0, interact: false, title: ""}, table: value :: Number
    row: 1
    row: 2
  end._rows-raw-array, 4)
```

should produce `"<Unknown value: details logged to console>"`

Now, try:

```
import plot-lib as P

_ :: Nothing = P.histogram({x-axis: "", y-axis: "", xscale: 0, yscale: 0, interact: false, title: ""}, table: value :: Number
    row: 1
    row: 2
  end._rows-raw-array, 4)
```

would print the above error instead.

This code is supposed to work. I don't know why it fails. It works fine on master branch (whose code is pretty much identical to horizon). I will investigate when I have time.

sorawee:
@blerner in `ffi.js`:

```
function throwTypeMismatch(val, typeName) {
      // NOTE(joe): can't use checkPyretVal here, because it will re-enter
      // this function and blow up... so bottom out at "nothing"
      if(!runtime.isPyretVal(val)) {
        val = runtime.namespace.get("nothing");
      }
      runtime.checkString(typeName);
      raise(err("generic-type-mismatch")(val, typeName));
}
```

schanzer:
This no longer produces an error. Closeable?

sorawee:
[...]

Issue #110, opened 2016-08-15
Get plots as `Image`s


jpolitz:
It would be useful to have a way to not just render plots, but also get their contents at a particular point as an `Image` for programmatic usage.

sorawee:
Merged.

Issue #109, opened 2016-08-15
Parameter to control the size of a bar chart


jpolitz:
Bar charts should come with parameters to control the size, (much like in #108).

I think width of bar and height of tallest bar might be two easy ways to give much of the necessary control.

Ping @blerner, who may also have useful context to add.

PR #120, opened 2016-08-27
+image +resize

sorawee:
@jpolitz @blerner

### DIFF
--- a/src/web/arr/trove/plot.arr
+++ b/src/web/arr/trove/plot.arr
@@ -14,7 +14,6 @@
 #
 #   display-multi-plot: display-multi-plot,
 #   default-options: default-options,
-#   default-window-options: default-window-options,
 #
 #   histogram: histogram,
 #   pie-chart: pie-chart,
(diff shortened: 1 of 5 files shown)


src/web/js/trove/d3-lib.js
```diff
@@ -397,70 +410,137 @@
...
+
+    var pauseStack;
+
+    if (RUNTIME.isNothing(restarter)) {
+      pauseStack = RUNTIME.pauseStack;
+    } else {
+      pauseStack = function(cb) { cb(restarter) };
+    }
+
+    pauseStack(function(restarter) {
```
sorawee: @jpolitz @blerner this is very hacky. It is to avoid calling `pauseStack` twice. However, each time a user clicks the resize button, it spawns a new stack frame. Recall that I should use `safeCall`?

blerner: I don't understand this code at all -- what are you trying to do here?

pauseStack is for use when you are in a JS function being called while on a Pyret stack, but you need to do some JS processing that might be asynchronous, for example, so you need to pause Pyret until it resumes.  Inside the function being passed to pauseStack, you're _not_ on a Pyret stack any more, and you have no stack-safety guarantees; it's all managed by you.  You get back into the Pyret stack by calling the restarter.

safeCall is when you're in a JS function on a Pyret stack, and you want to call something that may eventually call into Pyret and so might bounce the stack.  The two functions passed to safeCall are the things to do before making the potentially-deep call back into Pyret, and then things to do afterward with the result of that call.

runThunk is for use when you're _not_ on a Pyret stack, and you want to call into a Pyret function that might bounce the stack.  Its signature is similar to safeCall, though the value passed into the second function is a Result, not a Pyret value, and needs some unwrapping to distinguish if things succeeded or failed.

There is no fourth function, for JS calls that might call into JS code that's asynchronous -- that's just life with JS.

Given those four scenarios, what is your code trying to do, and can you state clearly what its assumptions are about the kind of stack it's running in, and the kinds of functions it's calling?

sorawee: I want to write the following function:

```
fun plot(window_size) {
  response-after-user-closes-window = open-plot-window()
  case (response-after-user-closes-window) {
    | user-says-exit => return image
    | user-says-resize => return plot(new-window-size)
  }
}
```

Things are complicated because 1) both opening the plot window and returning an image must be done asynchronously. Thus, `pauseStack` is essential. The simplified code is the following.

```
fun plot(window-size) {
  rt.pauseStack(fun(restarter) {
    open-plot-window(fun(response-after-user-closes-window) {
      case (response-after-user-closes-window) {
        | user-says-exit => 
          var image = new Image;
          image.onload = fu(){
            restarter.resume(image);
          }
        | user-says-resize => plot(new-window-size)
      }
    })
  });
}
```

The problem that I am having is, when I recursively call `plot`, it will hit `pauseStack` again. As you said, after I call `pauseStack`, I am no longer in Pyret stack, so calling it again will result in an error. What I did to avoid this is to write

```
fun plot(window-size, maybe-restarter) {
  my-pause-stack = case(maybe-restarter) {
    | none => rt.pauseStack
    | some(restarter) => fun(callback) { callback(restarter) }
  }
  my-pause-stack(fun(restarter) {
    open-plot-window(fun(response-after-user-closes-window) {
      case (response-after-user-closes-window) {
        | user-says-exit => 
          var image = new Image;
          image.onload = fu(){
            restarter.resume(image);
          }
        | user-says-resize => plot(new-window-size, some(restarter))
      }
    })
  });
}
```

But it's super hacky and might not be the right thing to do. And that's basically the question I'm asking. What's the right thing to do here?

sorawee:
This fixes #108, #109, #110, and #111

src/web/arr/trove/plot.arr
```diff
@@ -33,20 +32,27 @@ provide-types {
...
 import image-structs as I
+import image as IM
 include lists
 import plot-lib as P
 import either as E
 import string-dict as SD
 
 OFFSET = 1
-MAX-SAMPLES = 100000
+MAX_SAMPLES = 100000
```
jpolitz: Why lose the shish?

sorawee: My bad! I thought that's the convention given at https://cs.brown.edu/courses/cs019/2016/Pyret_Style_Guide.html, but I got it wrong. Will change back.

jpolitz:
I'm going to merge this to give it a whirl on horizon.  Would love to get this out ASAP.

I think we understand the stack situation now – do any code changes need to happen to resolve it?

(If we switch continuations to use, say `return`, as in another branch, this will require some changes, but so will other `pauseStack` using libraries.)

PR #122, opened 2016-09-02
Logging

jswrenn:
This extends #105 with:
 - a safer, simpler `render-fancy-reason` calling scheme and fallback protocol,
 - a basic means for logging events

The set of commits at the time of creating this pull request represent something of a minimal viable product. This is a work-in-progress. I'm creating this pull request to provide a place to track and discuss:
 - what to log,
 - how to log it

---
As of the latest commit, events are logged by `POST`ing them, as they are issued, to a url defined by the environment variable `LOG_URL`. There is also a `console.log` backend, and an incomplete `IndexedDB` backend. I welcome discussion of alternative strategies for syncing logs with the server.

Enough should be logged that it is possible to:
 - disambiguate between logged data from different tabs, sessions, users, etc.
 - perform an offline reconstruction of user interaction

Currently, these things are logged:
 - The running of definitions or REPL code
 - Document management `set` events (when a new name is associated with a new CodeMirror document)
 - CodeMirror edit events
 - Upon errors, the `torepr` of the error, the `torepr` of the `render-fancy-reason` or `render-reason` result
 - Highlight mode switches
 - Error message highlight activations
 - Highlight anchor mouseover and click
 - CodeMirror focus
 - Page visibility
 - User authentication

All events are additionally tagged with a:
 - timestamp
 - unique window ID number
 - unique browser session ID number
 - unique cross-session browser ID number


jpolitz:
```
> I welcome discussion of alternative strategies for syncing logs with the server.
```

I think it would be nice to put this at least somewhat under the user's control.  Put the logs in their Drive in a file somewhere, and then ask them to send these files to us, if they want to share them.

I think, regardless, we should avoid sending the requests directly to the servers of code.pyret.org.  If we do want to send them to a server under our control, we should fire up a separate CPO-logging service, and point the log requests there.  That could be a server hosted at Brown, another Heroku server, etc.  Its job is simply to aggregate the logs, and can be separately access controlled from the main server.

jswrenn:
> I think it would be nice to put this at least somewhat under the user's control. Put the logs in their Drive in a file somewhere, and then ask them to send these files to us, if they want to share them.

I like that idea too, but logging isn't tied to the user being logged in; it's stored as a browser setting.

> I think, regardless, we should avoid sending the requests directly to the servers of code.pyret.org.

Agree. It will need to be served over HTTPS to avoid mixed-content issues. I'm a little wary of hosting it at Brown since I've run into a few people (including myself) who've had issues with Brown's certs.

jpolitz:
It could be at Pyret-logging.herokuapp.com

What is the plan for ownership and management of the data?

src/web/js/log.js
```diff
@@ -22,9 +22,12 @@ var IndexedDBBackend = function() {
 
 var logger = (function(backend) {
   function guid() {
-    var array = new Uint32Array(4);
+    var array = new Uint32Array(6);
     window.crypto.getRandomValues(array);
-    return array;
+    return array.reduce(
```
jpolitz: This `reduce` method isn't defined in Safari on `Uint32Array`, which makes the whole page fail to load.

PR #125, opened 2016-09-19
Proxy some shared files through the server to avoid GAE issues

jpolitz:
This isn't done yet, but I'm making the PR so I can write down some notes about the design and folks can read the diff if they want.


jpolitz:
Overall design:

## On Publish/"Open" of an Image File

Client gives server the file id, and id of the share directory (easiest to get on the client, which already has this).

Using the requesting user's credential, the server creates a copy of the file, tries to make it publicly readable, and whether or not that succeeds, records an entry of the copied file's id and the current user's ID.

- If the copy fails, the whole request fails.  This ensures that the requesting user has permissions to make copies of the file they are trying to publish.  We don't check that the file is owned by the requesting user, just that they are able to copy it.
- The copied file is always owned by the requesting user and stored in their drive.
Since it's very difficult to test all the ways the "set to public" response can work (it appears to silently fail on student accounts), for now we unconditionally store the file/user mapping.  This is useful information to have around anyway, and it's fine to store information like this on a "Publish" action.

## On Metadata Fetch

There are two kinds of GDrive file fetching in CPO: one for the metadata of a file (used to show the filename, check for times the file has been shared, check modified times for caching, etc), and one for the content.  This is also how the Drive API is built.

Since we're now proxying that information through, but don't necessarily want to simply forward all the information about a file, we only return the information that is needed in the metadata request.  Here:

The client provides the id of a file (the thing after share=).  It first tries to fetch the file directly from Google; if the file is public or the user has permissions to it, this will succeed, and everything will proceed as normal.

If this fails (and only if it fails), the client makes a request to the server with the file's ID.  The server:

1. Looks up the file ID in its table of published files, and finds the corresponding user
2. Fetches an access token from Google for the stored user, using the stored refresh token for that user
3. Requests the file with the access token
4. Returns just the name, modified time, id, and a link back to CPO to download the contents of the file

## On Content Fetch

This does essentially the same thing as the metadata fetch, but does the appropriate Drive API request to get only the body of the file.  There are two endpoints for this:

- One for images, which returns the data in raw binary format, for use in img tags.  This errors if the file it fetches from Google doesn't have image as part of its MIME.
- One for programs, which returns the data as text/plain, for source code.  This errors if the file it fetches from Google doesn't have text/plain as its MIME.

PR #148, opened 2016-11-08
Add drive-view to list of IDs in the "More" menu

jas14:
The "View in Google Drive" link looked different. It was an underlined purple link instead of a block link like the rest of the links in the "More" section. This fixes that.


jpolitz:
Thanks!  This got incorporated and just pushed out; I just made the change along with some others I was working on at the time, so I'm not going to do the merge through here, but it does look right now!

PR #158, opened 2017-02-06
overlays were off because of rounding errors.

ds26gte:
Copies `OverlayImage` fix from wescheme2012 repo to CPO. 
Also adds the color `"TRANSPARENT"`.

### DIFF
--- a/src/web/js/trove/image-lib.js
+++ b/src/web/js/trove/image-lib.js
@@ -319,6 +319,7 @@
     colorDb.put("DIM GRAY", makeColor(105, 105, 105));
     colorDb.put("DIMGRAY", makeColor(105, 105, 105));
     colorDb.put("BLACK", makeColor(0, 0, 0));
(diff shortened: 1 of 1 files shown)


jpolitz:
Interesting.  I _added_ some of these `getWidth/getHeight` calls explicitly when porting over fixes from the WeScheme image library.

https://github.com/ds26gte/code.pyret.org/commit/65fac0c3c9adbdb9312f695ca1459999a1b41638#diff-437af9c73b7135d2af8be3d1d2b75c46R440

I'm having trouble paging in the history here.

schanzer:
My guess is that it *looked* like I wasn't using proper getter methods (which is true - but I failed to document why). I probably should have named those functions `getIntegerHeight()` and `getIntegerWidth()`

jpolitz:
@blerner you're listed for approval -- do you want to chime in?  It LGTM.

blerner:
I don't have the context paged in right now, so if both you and @schanzer approve, I'm not objecting.

Issue #164, opened 2017-02-24
Files should open in a new tab


kevinschaich:
(And use target to prevent repeated tabs when opening the file again)

This was the behavior we originally had but opted for same-tab instead. Do you think if we add a button in the IDE to "return to dashboard," it would be more intuitive to open in a new tab or same tab?

I think it just might be confusing if they have the dashboard open in another tab and then "return to dashboard" doesn't work when they click it.

schanzer:
"return to dashboard" should switch to the dashboard tab if it's already open, and open a new tab if it's not.

It's an extremely common use case to have the list of programs open in one tab, and a program you are editing in another.

kevinschaich:
Is this technically feasible? I'm not sure that modern browsers have a notion of "switch to this tab if it's open" via Javascript.

This could be further complicated if people have multiple browser windows open.

michaelmauer:
Also, I think that it might make sense to just open in a new tab and not bother taking control of how the tab changes if we close the IDE one. You don't always want to go back to the dashboard after closing the IDE, I think.

awstlaur:
Another option could be to implement custom tabs on the web app, akin to how a google sheets document has tabs for different sheets in the same document. Except the tabs would have separate documents. (I think @jswrenn mentioned this idea in person to me at one point).

shriram:
That seems like a huge implementation problem (cheap to ask, hard to build
*well*, taking dozens of UI issues into account).

schanzer:
This is not a js issue - HTML itself allows you to load pages in named windows. Easy to implement. 

(WeScheme has done this for years)

kevinschaich:
Pull request later today will resolve this, but I encountered some errors along the way with Chrome's popup blocker thinking that the "open IDE in a new tab" was spam.

Issue #163, opened 2017-02-24
Why is "Select from Drive" on the bottom right?


kevinschaich:
It has similar semantics to the "Recent Files" and "Templates" buttons - why is it in such a different place?

kevinschaich:
Will update to a separate tab "browse..." or similar to open file picker

kevinschaich:
Resolved in #165.

Issue #162, opened 2017-02-24
"New file" UI is awkward


kevinschaich:
Dorai:

> The text box on the bottom-left is really confusing. I initially thought it was a "live filter" control, which would allow me to sift through my saved files, then I thought that it was broken when nothing happened after typing.
> 
> In general, this workflow is confusing. If we're going to ask for a filename, we should do so from a modal dialog that pops up after the user clicks "New File". That way it's clean and unambiguous what the filename is for. Or, alternately, we could use GDrive's convention: make a new, untitled file and let the document editor handle the filename.

kevinschaich:
Will make this a button in the upper right -> opens a modal to input filename and confirm.

kevinschaich:
Resolved in #165.

Issue #160, opened 2017-02-24
Use Pyret icon for files in Student Dashboard


kevinschaich:

kevinschaich:
Fixed in #165.

PR #165, opened 2017-02-27
Update student dashboard code & move to new folder

kevinschaich:
Resolves the following issues: #162 (Awkward new file UI), #163 (move Select from Drive button), #164 (open Student Dashboard files in new tab), #160 (Use pyret logo for file icons).

Also refactors auth code to fix logins upon page refresh, and moves dashboard code into its own folders.

Also adds the `--watch` flag to the `webkit-dev-server` command, executed when `npm start` is run from the root directory. This allows files to hot-update and refresh themselves when using the development server.


src/web/js/dashboard/config.js
```diff
@@ -0,0 +1,9 @@
+// App name in *Title Case*.
+export const APP_NAME = 'Pyret';
```
jpolitz: I guess these should probably be centralized into the .env file (or some other central config location), right?

jpolitz:
In the interests of moving forward, I'm going to merge this; there's nothing bad or major in the changes, so any code comments I have don't need to be made now.  This will go up on pyret-horizon automatically in a little bit.

Issue #35, opened 2015-09-15
key events


schanzer:
This comes as no surprise to anyone, but I've been knee-deep in it with WeScheme for the last few days so I figured I might as well file a report. This is based on the Pyret program found at https://code.pyret.org/editor#share=0BzzMl1BJlJDkeFFTUVhURk1rd0E&v=v0.5r1428

1) Caps lock is ignored  (CapsLock+k -> k, instead of K)
2) Shift modifier is ignored  (Shift+k -> k, instead of K)
3) Alt/Opt modifier is ignored (Alt-c -> c, instead of ç)

I don't know how much of the old jsBigBang code Pyret inherited, but here's the issue: the only way to get modifier keys (shift, alt, etc) is to listen for keydown events, and the only [easy] way to get the modified character (ç) is to listen for keypress events. 

I briefly attempted to reconstruct the typed character solely from the keydown event, but that got complicated fast. For WeScheme, I instead signed up the key-event handler for both down *and* press, and then had it sift through some conditions to know when to ignore one and listen for the other. It feels like a bit of a fragile solution, but I don't expect anyone to re-assign the extended ASCII character set anytime soon. :) 

Hope this is helpful!

jpolitz:
Wait, if I do:

    document.body.addEventListener("keypress", function(e) { console.log(e); })

And type "Shift-K", I get an object with:
    
    altKey: false
    shiftKey: true
    metaKey: false
    ctrlKey: false

So some of these are certainly present.

If I type Option-[ (for curly quote on a Mac), I get:

    altKey: true
    shiftKey: false
    ...
    keyIdentifier: "U+005B" // unfortunately, thats [, not “, as we might hope

I think that the modifier keys are reasonably cross-browser (Caps Lock isn't considered a modifier key).  I think getting option-based input via OSX is probably best done by hiding a text area and focusing that, then reading the input out of it after an "input" event, rather than rebuilding the modifier logic ourselves.

schanzer:
Huh? keypress isn't fired if you just hit a modifier key.

jpolitz:
Agreed.  I didn't realize part of the goal was to fire key events for standalone modifier keys; I thought you were just saying it was hard to get the modifier information period.  I take it you want to support a key event that fires "Shift" _and_ having Shift+a show up as "A" for the key?  Or is there another reason keydown rather than keypress is necessary?

schanzer:
[...]

PR #167, opened 2017-04-20
Added keypress event for on_key to enable Unicode characters in world text

ds26gte:
Recognize `alt` and `ralt` keys, and have `on_key` treat `Compose` and `Backspace` as keydown and every other key as keypress. This allows world to respond to AltGr-generated Unicode characters (e.g., in text boxes).

(This is **not** related to `\b` in strings. `Backspace` continues to be treated as keydown as before. It's just all the other letters that are promoted to the new keypress.

It is also unrelated to allowing Unicode letters in Pyret ids.)

Here is an example to try it on, based on Tiina Partanen's WeScheme program:

```
import image as I
import world as W

var LEVEYS = 500
var KORKEUS = 50

var KUVA = I.empty-scene(LEVEYS, KORKEUS)

fun piirra(teksti):
  I.overlay-align("left", "middle", I.text(teksti, 20, "black"), KUVA)
end

fun kirjoita(teksti, nappain):
  if (string-length(teksti) > 0) and W.is-key-equal("backspace", nappain):
    string-substring(teksti, 0, string-length(teksti) - 1)
  else if W.is-key-equal("backspace", nappain):
    teksti
  else if string-length(nappain) == 1:
    string-append(teksti, nappain)
  else:
    teksti
  end
end

W.big-bang("", [list:
  W.to-draw(piirra),
  W.on-key(kirjoita)
  ])
```

### DIFF
--- a/src/web/js/trove/world-lib.js
+++ b/src/web/js/trove/world-lib.js
@@ -54,12 +54,6 @@
       return forEachHelp(0);
     };
 
-
-
-
-
-
-
     //
     // WORLD STUFFS
     //
@@ -76,8 +70,6 @@
     var worldListeners = null;
     var eventDetachers = null;
     var changingWorld = [];
-
-
 
     function clear_running_state() {
       worldIndexStack = [];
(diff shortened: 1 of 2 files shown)


jpolitz:
Closing this per new discussion at the end of https://github.com/brownplt/code.pyret.org/issues/35

PR #170, opened 2017-05-04
rotate takes arbitrary reals as angles and canonicalizes them internally to [0, 360)

ds26gte:
Fix for brownplt/pyret-lang#1014

### DIFF
--- a/src/web/js/trove/image.js
+++ b/src/web/js/trove/image.js
@@ -242,6 +242,18 @@
     var annAngle = ann("Angle (a number 'x' where 0 <= x < 360)", image.isAngle);
     var checkAngle = p(image.isAngle, "Angle");
 
+    var canonicalizeAngle = function(angle) {
+      angle = checkReal(angle);
+      while (jsnums.lessThan(angle, 0)) {
+        angle = jsnums.add(angle, 360);
+        //angle += 360;
+      }
+      while (jsnums.greaterThanOrEqual(angle, 360)) {
+        angle = jsnums.subtract(angle, 360);
+        //angle -= 360;
+      }
+      return angle;
+    };
 
     var annListColor = ann("List<Color>", function(val) {
       return runtime.ffi.isList(val);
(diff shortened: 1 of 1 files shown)


jpolitz:
LGTM

jpolitz:
Actually, I just tested this out, and I have a complaint.  This should probably not use a while loop.  The very first test I tried crashed the tab:

image

Can we just divide instead of having a hanging loop stuck in the middle of an innocuous image function?

Execution management is hard!

src/web/js/trove/image.js
```diff
@@ -242,6 +242,18 @@
     var annAngle = ann("Angle (a number 'x' where 0 <= x < 360)", image.isAngle);
     var checkAngle = p(image.isAngle, "Angle");
 
+    var canonicalizeAngle = function(angle) {
```
blerner: Not quite, that's wrong for negative angles (you'll be too small by 360 degrees).  But it's then an easy matter to fix up...

ds26gte: I'm thinking the following would be the most efficient in JS:

```
      angle = jsnums.remainder(angle, 360);
      if (jsnums.lessThan(angle, 0))
        angle = jsnums.add(angle, 360);
```

blerner: That looks fine, though what's the spec for remainder on negative arguments?  (I.e. is `remainder(-4, 3)` equal to -4 because remainder doesn't make sense for negatives, -1 because only one multiple of 3 fits into -4, 1 for unknown mathematical reasons, 2 because two multiples of 3 get to the largest number less than -4, or something else?)

ds26gte:
`jsnums.remainder(-4, 3)` is `-1`. Same semantics as Racket `remainder` and JS `%` (which is _not_ "modulo"). Also works for non-integral dividend (we need only worry about integral divisor 360).

blerner:
Great, then your code above looks fine :)

ds26gte:
Thanks. New PR at #172.

PR #171, opened 2017-05-05
Add plot lib to master

sorawee:
@SDooman @jpolitz

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "node-uuid": "^1.4.1",
     "pyret-codemirror-mode": "git://github.com/brownplt/pyret-codemirror-mode#master",
     "pyret-ide": "git://github.com/pcardune/pyret-ide#master",
-    "pyret-lang": "git://github.com/brownplt/pyret-lang.git#horizon",
+    "pyret-lang": "git://github.com/brownplt/pyret-lang.git#master",
     "q": "~1.4.1",
     "react": "^15.3.2",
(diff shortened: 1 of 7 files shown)


jpolitz:
Question @sorawee / @SDooman why did the names of functions need to change?

jpolitz:
If someone could add a commit that puts the old names back (and make sure they work), that would be useful – we broke some teachers' code with this change.

sorawee:
Recall that this is a feedback from @blerner. I can definitely change it back if that causes a problem.

blerner:
I don't remember saying that, but it sounds plausible...
Separately, does this include the fix for plot window margins that Jen requested for the physics usages?

jpolitz:
I totally believe that it makes sense, I just wanted the reason documented on the PR.  I don't think that we should change back, but we should leave the old names in.  I think just aliasing the new names to the old ones and providing both makes sense.

sorawee:
@blerner does not have the fix yet. Will do it today.

@jpolitz will alias the old name to the new name now.

sorawee:
I recall it now. Ben did not suggest changing this, but suggest changing `x-scale` to `extend-x`. Changing from `display-` to `render-` is mine, and the reason is that with `interact` setting to `false`, there's no display and the function just returns an image value, so naming it `display-` is misleading.

Anyway, I make aliases to `display-`. Will submit a PR soon.

PR #175, opened 2017-05-19
Update plot library

### LINKED ISSUE #168: Plot library's infer-bounds should be magnitude-aware
When using `infer-bounds` on a plot, the library computes the minimum and maximum x and y values, then expands their ranges by `OFFSET` in all four directions, where `OFFSET = 1`.  This leads to a lot of wasted space when the graph is of overall small-magnitude values.  (E.g., plotting `num-sin(x)` will use only the middle half of the graph area; plotting `num-sqr(num-sin(x))` will use only one-third of the available area.

I think a better heuristic would be, use `OFFSET = num-min((max-y - min-y) / 40, 1)`  The /40 quotient means that for small-range graphs you'll add 2.5% additional height above and below, for a total padding of 5%, while for large graphs (with a vertical range > 40), you'll add the same constant 1-unit padding you currently do.

sorawee:
- converting plot-list functions from accepting tables to lists
  (histogram, bar-chart, grouped-bar-chart)
- fix #168
- add display-* as aliases of render-* functions
- make bar chart display better (rotate text a little bit to avoid collision)

### DIFF
--- a/src/web/arr/trove/plot-list.arr
+++ b/src/web/arr/trove/plot-list.arr
@@ -175,38 +175,26 @@ fun check-base-window-options(options :: BaseWindowOptions) -> Nothing block:
 end
 
 sprintf = sprintf-maker()
+unsafe-equal = {(x :: Number, y :: Number): (x <= y) and (y <= x)}
 
 default-options = {<A>(x :: A): x}
 
-fun histogram(tab :: Table, n :: Number, options-generator :: WrappedHistogramWindowOptions) -> IM.Image block:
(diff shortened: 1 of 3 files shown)


sorawee:
@SDooman I have not done the following:

- Switch pie-chart in plot-list to accept lists
- Fix the drawing of pie chart when the text is long

jpolitz:
I didn't think of this consequence of just aliasing them:

image

I think this is fine, but it's interesting.

jpolitz:
Thanks for turning this around so quickly.

PR #177, opened 2017-06-08
Plot lib change for summer

sorawee:
@SDooman @jpolitz 

1. convert pie-chart to list style in plot-list
2. add checks preventing unequal length list for those that require the check
3. fix the bug when trivial data is provided (0 or 1 data point)
4. add freq-bar-chart

### DIFF
--- a/src/web/arr/trove/plot-list.arr
+++ b/src/web/arr/trove/plot-list.arr
@@ -38,11 +38,11 @@ import base as B
 import image-structs as I
 import image as IM
 include lists
+import sets as S
 import plot-lib as P
 import either as E
 import string-dict as SD
 
-OFFSET = 1
 MAX-SAMPLES = 100000
 
 type BaseWindowOptions = {
@@ -179,7 +179,11 @@ unsafe-equal = {(x :: Number, y :: Number): (x <= y) and (y <= x)}
 
 default-options = {<A>(x :: A): x}
 
-fun histogram(values :: List<Number>, n :: Number, options-generator :: WrappedHistogramWindowOptions) -> IM.Image block:
+fun histogram(
+    values :: List<Number>,
+    n :: Number,
+    options-generator :: WrappedHistogramWindowOptions
+) -> IM.Image block:
   doc: ```
        Consume a list of numbers, and a number of bins,
        and show a histogram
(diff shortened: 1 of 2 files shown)


src/web/arr/trove/plot-list.arr
```diff
@@ -197,83 +201,106 @@ where:
...
+    [list: 1, 2, 3],
+    [list: 1, 2, 3],
+    default-options) does-not-raise
 end
 
 fun bar-chart(
-    labels :: List<String>,
-    values :: List<Number>,
-    options-generator :: WrappedBarChartWindowOptions) -> IM.Image block:
+    label :: List<String>,
```
sorawee: @SDooman here

src/web/arr/trove/plot-list.arr
```diff
@@ -197,83 +201,106 @@ where:
...
-fun pie-chart(tab :: Table, options-generator :: WrappedPieChartWindowOptions) -> IM.Image block:
-  doc: 'Consume a table with two columns: `label` and `value`, and show a pie-chart'
-  when not(tab._header-raw-array =~ [raw-array: 'label', 'value']):
-    raise('pie-chart: expect a table with columns named `label` and `value`')
-  end
-  when raw-array-length(tab._rows-raw-array) == 0:
-    raise('pie-chart: expect the table to have at least one row')
-  end
+fun pie-chart(
+    label :: List<String>,
```
sorawee: and here @SDooman

SDooman:
looks great to me!

schanzer:
Whoever merges this: make sure to also close #985 when you do.

jpolitz:
If I'm reading this right, it changes function signatures.  We really ruined the day of a few physics teachers the last time we did that with plotting.  They're probably in summer now, but this will break their code from last year.

Is it intentional that this interface gets updated, and not feasible to introduce new functions with updated signatures?

jpolitz:
Also, is there a corresponding update on the docs repo?  I checked a few branches but don't see it.

sorawee:
@jpolitz the function signature change goes to the `plot-list.arr` file which is a new file and is right now not documented, so no one has used it, and it won't break anyone's code. The `plot.arr` is the one that physics teacher would have used, and that is left unchanged (in fact, that's precisely the reason why we create `plot-list.arr`).

Yes, I will need to do the documentation.

jpolitz:
Thanks for the explanation @sorawee – apologies for being overly cautious.

PR #178, opened 2017-06-13
render roughnum with trailing ellipsis

ds26gte:
Fix for brownplt/pyret-lang#664.

(This modifies the renderer for roughnums rather than changing `Roughnum.prototype.toString`, which breaks too many tests, because `toString` isn't called only by the renderer.)

### DIFF
--- a/src/web/js/output-ui.js
+++ b/src/web/js/output-ui.js
@@ -1126,6 +1126,8 @@
           });
 
           return outText;
+        } else if (jsnums.isRoughnum(num)) {
+          return renderText(sooper(renderers, "number", num.toString() + '...'));
         } else {
           return renderText(sooper(renderers, "number", num));
         }


blerner:
This also seems like the wrong place to solve this.  Either Pyret's rendering behavior should include ..., or it shouldn't;  I don't think CPO and cmdline-Pyret should differ here.

ds26gte:
You mean you'd like the `...` in the default renderers in `pyret-lang/**/runtime.js`, as opposed to in `code.pyret.org/**/output-ui.js`?

The latter seems to already be making rendering decisions, e.g., the rational-number toggle. (I've also been adding dialect-conditional rendering decisions there, although they will only show in the Patch-specific fork of `code.pyret.org` and not here.)

In any case, changing `Roughnum.*.toString()` is too early (I tried it). The fix has to be made at the rendering stage, regardless of whether in `runtime.js` along with the default renderers, or in `output-ui.js` with the overriding renderers.

jpolitz:
Hm.  I'm really not happy with either choice, because I don't know what principle is being applied here.  The inconsistency between command-line and browser sucks, but it's also annoying to have another printed form that isn't re-parseable as the value.

I'm less unhappy with having it be consistent between command line and browser, so I agree with Ben.  I'd be more happy if we had a rule that demanded we go one way or another.

blerner:
We technically _could_ make the lexer's roughnum syntax end in an _optional_ `...`, so that the printed form is a subset of the lexable form.  That seems hacky, too, but is less bad than violating our goal of having `torepr` produce an expression that should evaluate back to "the original" value.

ds26gte:
Having the trailing-ellipsis form reparseable as a roughnum is not at all a big deal. (Requires adding an optional dot-dot-dot suffix to the roughnum regexps in `pyret-tokenizer` and `js-numbers`.) If that would satisfy you, let me know.

PR #179, opened 2017-06-16
stop-when takes an optional last-picture-handler argument

ds26gte:
Like WeScheme does. The last-picture-handler argument takes the final value of the reactor's state and returns an image, which is displayed as the last picture.

### DIFF
--- a/src/web/js/trove/world-lib.js
+++ b/src/web/js/trove/world-lib.js
@@ -749,7 +749,19 @@
               if (!stop) {
                 k2();
               } else {
-                if (extras.closeWhenStop) {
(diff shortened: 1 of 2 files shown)


blerner:
This is intended to address https://github.com/brownplt/pyret-lang/issues/808, correct?  If so, the design there was to add a new handler, `last-picture`, that took this optional last_picture_handler.  We don't want variable-arity functions in Pyret, by design.  In fact, the code as written gives the wrong arity error ("Expected 1 argument", but it should say "Expected 1 or 2 arguments" according to how this code is implemented).

ds26gte:
Thanks; will re-do to match Pyret policy.

PR #180, opened 2017-06-16
Added last-picture option for big-bang.

ds26gte:
Inspired by WeScheme's `stop-when`'s optional last-picture argument. However, since Pyret prefers fixed-arity procedures, this optional argument is supplied via a dedicated option called `last-picture`.

`last-picture` was also discussed in relation to Pyret in brownplt/pyret-lang#808.

### DIFF
--- a/src/web/js/trove/world-lib.js
+++ b/src/web/js/trove/world-lib.js
@@ -656,7 +656,7 @@
 
     //////////////////////////////////////////////////////////////////////
 
-    var bigBang, StopWhenHandler;
+    var bigBang, StopWhenHandler, LastPictureHandler;
 
     function BigBangRecord(top, world, handlerCreators, handlers, attribs,
                            success, fail) {
@@ -670,19 +670,23 @@
     }
 
     BigBangRecord.prototype.restart = function() {
-      var i;
+      var i, handler;
       for(i = 0 ; i < this.handlers.length; i++) {
-        if (! (this.handlers[i] instanceof StopWhenHandler)) {
-          this.handlers[i].onRegister(this.top);
+        handler = this.handlers[i];
+        if ((! (handler instanceof StopWhenHandler)) &&
+            (! (handler instanceof LastPictureHandler))) {
+          handler.onRegister(this.top);
         }
       }
     };
 
     BigBangRecord.prototype.pause = function() {
-      var i;
+      var i, handler;
       for(i = 0 ; i < this.handlers.length; i++) {
-        if (! (this.handlers[i] instanceof StopWhenHandler)) {
-          this.handlers[i].onUnregister(this.top);
+        handler = this.handlers[i];
+        if ((! (handler instanceof StopWhenHandler)) &&
+            (! (handler instanceof LastPictureHandler))) {
+          handler.onUnregister(this.top);
         }
       }
     };
@@ -736,9 +740,14 @@
       // Monitor for termination and register the other handlers.
       var stopWhen = new StopWhenHandler(function(w, k2) { k2(false); },
                                          function(w, k2) { k2(w); });
+      var lpf;
+      var handler;
       for(i = 0 ; i < handlers.length; i++) {
-        if (handlers[i] instanceof StopWhenHandler) {
-          stopWhen = handlers[i];
+        handler = handlers[i];
+        if (handler instanceof StopWhenHandler) {
+          stopWhen = handler;
+        } else if (handler instanceof LastPictureHandler) {
+          lpf = handler.lastPictureFunction;
         }
       }
       activationRecord.restart();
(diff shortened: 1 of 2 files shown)


shriram:
Thanks, but I'm not sure I'm a fan of this name. How about something like
last-image? "picture" is not a word we commonly use in world programming.

blerner:
For now, it should be a "scene", not an "image", right?

shriram:
I like that even more.

schanzer:
Didn't we talk about phasing out the distinction between Scene and Image a while back, since Scene is a holdover from Racket? We've already removed them from BS:A, and have no plans to mention them in Reactive. Why not just use Image?

ds26gte:
"Scene" is a continuous sequence of images, no?, as opposed to a static tableau?

blerner:
No, there's nothing about sequencing of images involved.  In my terminology in Fundies 2, when I do this in Java, Images are things that are placed together relatively (i.e., above, beside, overly-xy, etc), whereas Scenes can have images placed onto them at coordinates (place-image-xy).  They're canvases onto which images can be drawn, in some sense.

If we switch the image library over to the API I've proposed elsewhere, then the distinction mostly goes away, since all the various origins can get explicitly referenced, and we can get away with eliminating Scenes.

ds26gte:
OK. I was being dramatic. (Act 4, Scene 2, and the like.)

schanzer:
> If we switch the image library over to the API I've proposed elsewhere, then the distinction mostly goes away, since all the various origins can get explicitly referenced, and we can get away with eliminating Scenes.

So shouldn't this be `last-image`, as opposed to `last-scene`?

schanzer:
> If we switch the image library over to the API I've proposed elsewhere, then the distinction mostly goes away, since all the various origins can get explicitly referenced, and we can get away with eliminating Scenes.

So shouldn't this be `last-image`, as opposed to `last-scene`?

blerner:
At that point, yes.  Until then, no :)

schanzer:
Sorry, I can't tell if the smiley face makes this a joke or not. I'd _really_ like to avoid changing the API once it's in place. If we know it'll ultimately be `last-image`, and scenes are merely subtly different from images, can we call this `last-image` now?

blerner:
No, I wasn't joking.  We _know_ we're going to be changing the API -- you said you wanted the new image API, and you said you don't want it this summer since you're already in teacher-training mode with Emma and don't want to switch right now.  That's fair; so I was trying to optimize the API so that the concepts matched.  But, since our current library is fairly schizophrenic when it comes to scenes and images, I just don't care what we call this, since none of it matters right now -- any choice is inconsistent and is going to change somehow when the new image API lands.

schanzer:
Then if you don't care, I'm all in for last-image.

blerner:
Ok, @ds26gte , please rename to `last-image`...

src/web/js/trove/world-lib.js
```diff
@@ -749,7 +758,18 @@
...
               } else {
-                if (extras.closeWhenStop) {
+                if (lpf) {
+                  var handler = lpf();
+                  var handler1 = handler(thisWorldIndex);
+                  handler1.onRegister(top);
+                  handler1._listener(w, oldW, function(v) { k2(); });
+                }
+
+                if (lpf) {
```
jpolitz: This appears to be a fork in the control flow.  I think this should do `activationrecord.pause()`, and then call `handler1._listener` as above, and that should be the only control flow in this case.  The `closeWhenStop` and `else` cases should be unchanged, and they should all be in tail position.

src/web/js/trove/world.js
```diff
@@ -575,10 +545,50 @@
...
+      WorldConfigOption.call(this, 'last-image');
+      this.handler = handler;
+    };
+
+    LastPicture.prototype = Object.create(WorldConfigOption.prototype);
+
+    LastPicture.prototype.toRawHandler = function(toplevelNode) {
+      var that = this;
+      var reusableCanvas;
+      var lastPictureFunction = function() {
```
jpolitz: It would be nice if the pieces of this that are copy-pasted from the draw handler could be abstracted

src/web/js/trove/world.js
```diff
@@ -205,7 +210,7 @@
 
       var configs = [];
       var isOutputConfigSeen = false;
-      var closeWhenStop = false;
+      var closeWhenStop = true; //false;
```
jpolitz: I think this default should remain `false`.

src/web/js/trove/world.js
ds26gte: Added PR #191 that puts in all these suggestions and supersedes this PR. (Used new PR so as keep track of changed `horizon`.)

schanzer:
Subsumed by #218

schanzer:
Paging @ds26gte

schanzer:
Closing a PR doesn't remove the history, and this is subsumed by two _other_ PRs. Closing.

Issue #181, opened 2017-06-16
empty-scene should take (background) color argument?


ds26gte:
Pyret's `empty-scene` currently takes two arguments: width and height. WeScheme's `empty-scene` takes a third, optional, argument for (background) color. 

Should we add a third (non-optional) argument to Pyret's `empty-scene`, or have a separate procedure that takes the third argument?

shriram:
I prefer a separate one. Let's keep the default operators simple.

schanzer:
Especially if scenes are ultimately going away (see new image API)

ds26gte:
What's a good name for this procedure? `empty-color-scene`, `empty-colored-scene`, `empty-scene-color[ed]`, `empty-colored-background`, ...?

schanzer:
`empty-color-scene` works for me. No BS courses use it, and it's almost certain to be deprecated quickly.

blerner:
Available in horizon.

PR #182, opened 2017-06-18
empty-color-scene

ds26gte:
Fix for #181.

`empty-color-scene(20, 20, "purple")` returns an empty 20x20 scene with a purple background.

### DIFF
--- a/src/web/js/trove/image-lib.js
+++ b/src/web/js/trove/image-lib.js
@@ -319,6 +319,7 @@
     colorDb.put("DIM GRAY", makeColor(105, 105, 105));
     colorDb.put("DIMGRAY", makeColor(105, 105, 105));
     colorDb.put("BLACK", makeColor(0, 0, 0));
+    colorDb.put("TRANSPARENT", makeColor(0, 0, 0, 0));
 
     // clone: object -> object
     // Copies an object.  The new object should respond like the old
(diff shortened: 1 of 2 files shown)


src/web/js/trove/image-lib.js
```diff
@@ -766,15 +768,17 @@
...
     };
 
     // render: 2d-context primitive-number primitive-number -> void
     SceneImage.prototype.render = function(ctx, x, y) {
       var childImage, childX, childY;
       // create a clipping region around the boundaries of the Scene
       ctx.save();
-      ctx.fillStyle = "rgba(0,0,0,0)";
+      //ctx.fillStyle = "rgba(0,0,0,0)";
+      ctx.fillStyle = this.color? colorString(this.color) : 'transparent';
```
blerner: Why this conditional?  Why not revise `empty-scene` to pass in the transparent color explicitly, and never worry about undefined values flowing in here at all?

ds26gte: `empty-[color-]scene` are not the only procedures that create `SceneImage`s. The latter needs some way to know what `fillStyle` to use when it comes time to render.

blerner:
I'm going to fold this into brownplt/pyret-lang#417, and close this PR itself.

PR #184, opened 2017-06-24
fix the histogram bug & move the label outside for pie-chart

sorawee:
The title describes everything in this change.

@jpolitz @SDooman

### DIFF
--- a/src/web/js/trove/plot-lib.js
+++ b/src/web/js/trove/plot-lib.js
@@ -910,14 +910,14 @@
 
 
     var dimension = getDimension({
-      minWindowWidth: 600,
-      minWindowHeight: 400,
-      outerMarginLeft: 160,
-      outerMarginRight: 160,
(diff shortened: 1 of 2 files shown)


src/web/js/trove/plot-lib.js
```diff
@@ -910,14 +910,14 @@
...
-      marginTop: 50,
-      marginBottom: 10,
+      minWindowWidth: 700,
+      minWindowHeight: 550,
+      outerMarginLeft: 10,
+      outerMarginRight: 10,
+      marginLeft: 120,
+      marginRight: 120,
+      marginTop: 90,
+      marginBottom: 40,
```
jpolitz: Comments on why these magic numbers changed?

sorawee: I move labels outside of the chart to avoid collision, but that means the chart will take more space

PR #190, opened 2017-06-30
Fix https://github.com/brownplt/pyret-lang/issues/262

blerner:


sorawee:
Just want to make sure: this is a breaking change, correct?

blerner:
Yes.  I've already fixed the relevant bits of d3-lib and plot-lib and image-lib to match, so no one using those should notice.  Only if you were manually creating a color that wasn't in the image-structs library would you notice any difference.

PR #209, opened 2017-08-03
makeScaleImage: flip image appropriately when xFactor and/or yFactor are negative

ds26gte:
Fix for brownplt/pyret-lang#1122.

Here's a test you can try: negative xFactor flips right<->left; negative yFactor up<->down.

```
include image

lion = bitmap-url('http://ds26gte.github.io/tex2page/t2p.png')

lion-left = scale-xy(-1, 1, lion)

lion-down = scale-xy(1, -1, lion)
```

### DIFF
--- a/src/web/js/trove/image-lib.js
+++ b/src/web/js/trove/image-lib.js
@@ -1177,8 +1177,8 @@
       });
 
       this.img      = img;
-      this.width    = img.width * xFactor;
-      this.height   = img.height * yFactor;
+      this.width    = img.width * Math.abs(xFactor);
+      this.height   = img.height * Math.abs(yFactor);
       this.xFactor  = xFactor;
       this.yFactor  = yFactor;
       this.ariaText = "Scaled Image, "+ (xFactor===yFactor? "by "+xFactor
@@ -1193,6 +1193,12 @@
     ScaleImage.prototype.render = function(ctx, x, y) {
       ctx.save();
       ctx.scale(this.xFactor, this.yFactor);
+      if (this.xFactor < 0) {
+        ctx.translate((this.width+2*x)/(this.xFactor), 0);
+      }
+      if (this.yFactor < 0) {
+        ctx.translate(0, (this.height+2*y)/(this.yFactor));
+      }
       this.img.render(ctx, x / this.xFactor, y / this.yFactor);
       ctx.restore();
     };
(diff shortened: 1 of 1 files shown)


blerner:
This seems like the expedient and yet somehow _wrong_ thing to do: why is it necessary to negate the scale factor and use a flip image?  Said another way, when using negative factors, what computations go wrong in the bounding-box calculations that cause the image not to appear correctly?  Need a comment at least, to explain why this is necessary as opposed to the obvious behavior as scale factors shrink to negative numbers?

ds26gte:
SInce `(0,0)` is the top-left corner, a negatively scaled image goes into an invisible quadrant. It _will_ be visible if we overlay it with appropriate offset on a larger image. The original state of affairs may be desirable after all. (Explicitly flipping an image flips it about its center, so it remains visible.)

We could I guess include the axes as a grayish cross-hairs that moves down and right just the right amount as non-blank images start occupying anything but the positive quadrant...?

blerner:
No, I definitely want negative scale factors to result in a flipped image: it's the logical extension of how to interpret the scale factor, and leads to a fun and useful teaching moment about graphics, about domain/range misconceptions, about testing, and other fun stuff.  I guess my surprise came because I thought that every image operation implicitly recentered the resulting image to its local origin, so the fact that it was rendering at negative pixel locations wouldn't matter.

We *definitely* do not want to start drawing axes on student's pictures -- that's extremely disconcerting ("I didn't tell it to do that!") and I routinely give assignments where the most convenient/logical way to build up the image involves moving things outside quadrant I.

ds26gte:
OK, updated my PR.  Negative scaling results in visible images, with the underlying canvas being shifted for convenience, to show the usually hidden quadrants. (This is somewhat similar to what 'flip' does too, but I don't have 'scale' call 'flip' anymore.) Some extra arithmetic was added to ensure that the scaled image's width & height are always positive. Because of the way the canvas shifting works (addition followed by division), it was not possible to have negative width & height be logically tenable, which I think is ok. However, the sign of the xFactor & yFactor are not messed with, so the "natural" origin of the image is not lost, so you should be able to indefinitely iterate the same negative scaling without the image marching off across the canvas.

schanzer:
Any reason this can't be merged?

blerner:
I'm ok with it

schanzer:
Is this still merge-able? Or have things changes sufficiently that it’s obsolete?

blerner:
I need to double-check this against the pinhole stuff I have coming in the reworked image library, to make sure that my pinholes account for the flipping properly.  Then it's probably fine.

blerner:
Ok. I looked into this today: this fix is partial, and needs some other updates to work with pinholes properly.  I'll incorporate these changes directly into https://github.com/brownplt/pyret-lang/issues/417, so this PR is probably obsolete now and can be closed.

PR #225, opened 2017-08-31
plot library using google charts

### LINKED ISSUE #108: Parameter to control the radius of a pie chart

sorawee:
Functionalities added (compared to the old library)
- introduce a namespace `from-list` (so that when we have a table interface, we can add `from-table`)
  That is, write `from-list.pie-chart` instead of `pie-chart`.
- resizing window will resize the chart automatically (effectively fix #108 and #109, I think?)
- labeled-histogram
- labeled-scatter-plot (brownplt/pyret-lang#1097)
- Series -> DataSeries (brownplt/pyret-lang#1091)
- combine y-min and y-max to y-range, x-min and x-max to x-range to save typing, have a nice
  default value and improve error checking. Plus, this allows a new feature -- set `x-range` for function-plot, and it will compute `y-range` automatically.
- exploding pie chart (though adjustable pie chart is removed because Google does not support)

Issues that are fixed or no longer in our control:
- Labels for zero on scatterplot axes look like points (brownplt/pyret-lang#1090)
- bar chart not using full width (brownplt/pyret-lang#1087)
- pie chart needs to show more info (brownplt/pyret-lang#1079)
- frequency chart needs to use integers (brownplt/pyret-lang#1080)
- y-max method not implemented on bar charts in plot-list library (brownplt/pyret-lang#1123)


src/web/arr/trove/chart.arr
```diff
@@ -0,0 +1,1028 @@
...
+              .map(_.get-data())
+          | some(shadow function-plots-data) => function-plots-data
+        end
+
+        scatters-arr = for map(p from scatter-plots + function-plots-data):
+          ps-to-arr(p.{ps: p.ps.filter(in-bound-xy(_, self))})
+        end ^ reverse ^ builtins.list-to-raw-array
+
+        # NOTE: lines-arr has JavaScript's null in it.
+        # Don't try to inspect it from Pyret's side
```
blerner: WAT?  This sounds very very brittle to me...

src/web/arr/trove/chart.arr
```diff
@@ -0,0 +1,1028 @@
...
+        shadow function-plots-data = cases (Option) function-plots-data:
+          | none => function-plots
+              .map(generate-xy(_, self.x-min, self.x-max, self.num-samples))
+              .map(_.get-data())
+          | some(shadow function-plots-data) => function-plots-data
+        end
+
+        scatters-arr = for map(p from scatter-plots + function-plots-data):
+          ps-to-arr(p.{ps: p.ps.filter(in-bound-xy(_, self))})
+        end ^ reverse ^ builtins.list-to-raw-array
```
blerner: As an efficiency thing here, you should probably drop down to arrays immediately, rather than create 3 intermediate arrays (4 if you count function-plots-data itself) before dropping down.  You could do

```
scatters-arr-rev = raw-array-concat(list-to-raw-array(scatter-plots), function-plots-data-arr) # assuming this was built as an array already
scatters-arr-len = raw-array-length(scatters-arr-rev)
scatters-arr = for raw-array-build(idx from scatters-arr-len):
  p = raw-array-get(scatters-arr-rev, scatters-arr-len - idx - 1)
  ps-to-arr(p.{ps: p.ps.filter(in-bound-xy(_, self))})
end
```
and similarly elsewhere.

src/web/js/output-ui.js
```diff
@@ -1275,10 +1275,11 @@
           var originalImageDom = img.toDomNode();
           $(container).click(function(e) {
             var dialog = $("<div>");
+            // NOTE(Oak): some magic numbers that "display" nicely
```
blerner: Ugh, those are seriously magic numbers.  Where do they come from?  Do they work on all reasonable screen resolutions and sizes?  (Especially projected screens, and laptops...)

src/web/js/trove/google-charts.js
```diff
@@ -0,0 +1,226 @@
+(function(){var a="\n//# sourceURL=",k="' of type ",n='<script type="text/javascript" src="',p="SCRIPT",r="array",t="complete",u="function",v="google.charts.load",w="hasOwnProperty",x="number",y="object",z="pre-45",A="propertyIsEnumerable",B="string",C="text/javascript",D="toLocaleString";function E(){return function(b){return b}}function F(){return function(){}}function G(b){return function(){return this[b]}}var I,J=J||{};J.scope={};
```
blerner: Yikes -- do we need to commit this to our repo?  Is there an npm package we could use instead?

src/web/arr/trove/chart.arr
```diff
@@ -0,0 +1,1028 @@
...
+}
+
+################################################################################
+# DATA DEFINITIONS
+################################################################################
+
+data DataSeries:
+  | line-plot-series(obj :: LinePlotSeries) with:
+    is-single: false,
+    constr: {(): line-plot-series},
```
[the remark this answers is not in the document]
sorawee: because without it being thunk, they are unbound-id.

src/web/arr/trove/chart.arr
```diff
@@ -0,0 +1,1028 @@
...
+    width: width-method,
+    height: height-method,
+  | bar-chart-window(obj :: ChartWindowObject) with:
+    constr: {(): bar-chart-window},
+    title: title-method,
+    width: width-method,
+    height: height-method,
+    x-axis: x-axis-method,
+    y-axis: y-axis-method,
+    y-range: y-range-method,
```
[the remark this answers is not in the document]
sorawee: I'm not exactly sure what is your concern, so let me put all of my thought when I designed this, and hopefully it will address your concern.

1. This is bar-chart, so it can't have `x-range` (if that's your concern)
2. Having both `y-min` and `y-max` 0 will automatically compute the bound for y-axis. Initially, I have `y-min` and `y-max` separately, but then there is no good value for "please compute the bound for me automatically" because 0 could actually mean value 0. But if we combine them to `y-range`, we know that the range [0, 0] doesn't make sense, so we can use that as a default value for "please infer it for me".
3. I don't want to have `range(x-min, x-max, y-min, y-max)` because it's not easy to remember -- is it `range(x-min, x-max, y-min, y-max)` or `range(y-min, y-max, x-min, x-max)` or `range(x-min, y-min, x-max, y-max)`?
4. I want to be able to have a way to "reset" back. For example, I could write:

```
f(plot(...)
  .y-range(..., ...))
```

Whoever (not necessary the one who writes the above code) writing function `f` might want to set `y-range` to the default value and then display it. This interface provides a way to do this nicely.

src/web/js/trove/google-charts.js
sorawee: My bad! This can be gone. I include it in `editor.html` instead.

src/web/arr/trove/chart.arr
sorawee: Yup, sorry. This is really just a hack for now (since @schanzer seems to want the library to finish soon). It will be gone eventually when I have time to rework this part.

src/web/js/output-ui.js
sorawee: It's the width / height of, say, title bar of jQuery dialog. You have a good point, since I have not tested on other browser besides Firefox and Chrome, but it looks OK on those two.

src/web/arr/trove/chart.arr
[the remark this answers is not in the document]
blerner: 1. No, not my worry
2. That's what I'm worried about -- I'd like a way to say "I only want to specify the max (or min, respectively); please automatically compute the other one"
3. No, I'm not saying that either.  I'm saying having `y-min`, `y-max`, `x-min` and `x-max` as four separate methods gives finer grained control.
4. That's fine, but that's an orthogonal concern to what I'm suggesting.

In some sense, specifying `y-min = +Infinity` is a valid reset for the minimum value, and `y-max = -Infinity` is a valid reset for the maximum.  You could store those values internally as the defaults, then use the four individual methods I suggest above to override them, and then your `reset` method could set things back to the infinities that are the defaults.

src/web/arr/trove/chart.arr
blerner: This really worries me; brittle hacks have an annoying tendency to stick around longer than we like.

src/web/js/output-ui.js
blerner: At least leave a comment explaining where you guesstimated the numbers from, otherwise we'll never be able to reconstruct this logic.  Why are you specifying the numbers here anyway, rather than nesting a div one layer deeper, and specifying the actual, correct numbers on that inner div, then let the outer one take care of itself?

schanzer:
@sorawee - confirming that this PR does _not_ modify the rendering of a `DataSeries`?

sorawee:
What do you mean by modifying rendering?

schanzer:
`draw-plot("some title", some-series)` will use the new chart API, right? I'm asking about simply evaluating `some-series` in the REPL. Right now it prints out `histogram-like-series(...)`, and we'd talked about making that rendering appear more concrete for users. I'm asking if this PR includes that change as well, and my understanding is that it does not.

sorawee:
Ah, it changes, but you just pointed out that I forgot to add `...` to the end of the list. Will fix.

screenshot from 2017-09-04 16-30-52

src/web/arr/trove/chart.arr
```diff
@@ -0,0 +1,1028 @@
...
+type Posn = RawArray<Number>
+type TableIntern = RawArray<RawArray<Any>>
+
+################################################################################
+# HELPERS
+################################################################################
+
+fst = raw-array-get(_, 0)
+snd = raw-array-get(_, 1)
+posn = {(x :: Number, y :: Number, z :: String): [raw-array: x, y, z]}
```
jpolitz: Cool to see a use of raw arrays for this. I think this is the first library that's really built this way with perf in mind. Does this make interop with the JS easier, too?

src/web/arr/trove/chart.arr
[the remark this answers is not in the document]
jpolitz: That's an interesting design point for how `data` blocks desugar. Right now the values in the `with:` block are letrec-bound before the constructor is created, so you get a dynamic “this name not initialized yet”; this is an example where it would be better if that weren't true. But that would require an extra bit of state to create the constructor, then create its `with:` fields, then tie the loop cycle later to make sure the constructed objects can refer to these fields.

@justinpombrio @shriram this is an interesting scope desugaring case. I don't recall making an explicit design decision here, and I think this just fell out of the “obvious” implementation.

src/web/arr/trove/chart.arr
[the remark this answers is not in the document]
jpolitz: Note that infinity isn't a value in Pyret.

I agree with Ben's suggesting of multiple methods. Why not start auto-computed for both, and then have the `y-max` and `y-min` methods cause them to be individually set?

src/web/arr/trove/chart.arr
jpolitz: Agree with Ben. “There's nothing quite as permanent as a temporary fix”

src/web/js/trove/google-charts.js
jpolitz: If it's just import statements, I wonder if we could make `npm-dependencies` do the right thing. That runs through `browserify`, which should be able to handle import statements. I wonder if that file could have an `import` statement in it without breaking anything.

jpolitz:
This mostly LGTM. I would like to fix the nulls-in-array issue, and have a final answer on how the module gets included.

The other issues are more minor and can wait.

src/web/arr/trove/chart.arr
sorawee: There are two variants: `from-list` which we have now, and `from-table` which will be added after the new table interface is really finalized. Both variants construct the same `DataSeries`, and I hope that its `_output` will be table based.

I use `raw-array` as default because 1) that's what table is (underlying), so when we decide to implement fancy `_output`, it will be easier and 2) so that JS side needs not to worry about lists, though this is not really an issue.

src/scripts/npm-dependencies.js
```diff
@@ -31,6 +31,8 @@ define("d3", [], function() { return d3; });
 d3_tip = require("d3-tip");
 define("d3-tip", [], function() { return d3_tip; });
 
+//google_charts = require("../../src/web/js/trove/google-charts");
+//define("google-charts", [], function() { return google_charts; });
```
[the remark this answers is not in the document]
sorawee: So I can still load it through html, right? I just need to change the code to imitate Q.

src/web/arr/trove/chart.arr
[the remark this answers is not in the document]
blerner: I'm pretty sure this has bitten us in the past before.  It's challenging, since we want to have a shared `base` object for every instance of a given constructor, but I think we need to split the fields that are methods from fields that aren't (which are syntactically easily distinguishable), and only the methods go on the shared base object.  Then, I think a safe but ugly desugaring for this would be
```js
function makeConstr(arg1, arg2, arg3) {
  var dict = runtime.create(baseConstr);
  dict["arg1Name"] = arg1; dict["arg2Name"] = arg2; dict["arg3Name"] = arg3;
  var ans = new Construct(dict, brands);
  dict["withField1"] = computeWithField1(ans); // gotta thunk it
  return ans;
}
```

src/scripts/npm-dependencies.js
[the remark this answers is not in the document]
jpolitz: Yeah – I just want to make dependencies in Pyret-facing modules explicit.

sorawee:
@jpolitz please review

sorawee:
Let me merge it now so that Emmanuel can play with it. All major issues that Joe suggested to fix were fixed, so this should be OK.

jpolitz:
Yeah this looks good to me in a quick review; I suspect we're likely to get more useful feedback from Emu using it than anything else.

PR #232, opened 2017-09-15
fixes #231

### LINKED ISSUE #231: Support Apple shortcut keys
Ctrl-S -> Cmd-S on MacOS

YanyanR:
All the "Ctrl" keyboard shortcut now also works with "Cmd", e.g. both "Ctrl-s" and "Cmd-s" saves the file. @schanzer

### DIFF
--- a/src/web/js/cpo-main.js
+++ b/src/web/js/cpo-main.js
@@ -531,23 +531,23 @@
       }
 
       $(window).on("keydown", function(e) {
-        if(e.ctrlKey) {
-          if(e.keyCode === 83) { // "Ctrl-s"
+        if(e.ctrlKey || e.metaKey) { // Ctrl or Cmd
+          if(e.keyCode === 83) { // "Ctrl-s" or "Cmd-s" to save
             CPO.save();
             e.stopImmediatePropagation();
             e.preventDefault();
           }
-          else if(e.keyCode === 77) { // "Ctrl-m"
+          else if(e.keyCode === 77) { // "Ctrl-m" or "Cmd-m" to resize REPL width to default
             toggleEditorSize();
             e.stopImmediatePropagation();
             e.preventDefault();
           }
-          else if(e.keyCode === 13) { // "Ctrl-Enter"
+          else if(e.keyCode === 13) { // "Ctrl-Enter" or "Cmd-Enter" to run code
             doRunAction(editor.cm.getValue());
             CPO.autoSave();
             e.stopImmediatePropagation();
             e.preventDefault();
-          } else if(e.keyCode === 191 && e.shiftKey) { // "Ctrl-?"
+          } else if(e.keyCode === 191 && e.shiftKey) { // "Ctrl-?" to pull up help menu
             $("#help-keys").fadeIn(100);
             e.stopImmediatePropagation();
             e.preventDefault();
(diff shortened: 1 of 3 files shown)


blerner:
This needs to be tested on Windows machines too, on Chrome and Firefox: the meta key is *sometimes* mapped to the Windows key instead (and I can't find reliable information online about it right now), which isn't what would be expected on a Windows machine.

YanyanR:
On Windows keyboards, this is the windows key (⊞). (source: https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/metaKey)

But I think this is fine, if we support both. 
On windows computer, they can still use "Ctrl-s" to save. Even if someone accidentally use "Windows-s", then that's also fine. On Mac, default "Cmd-s" will save the html file, but I don't think anyone would have use for that.

blerner:
No, on Windows, the Windows+S combination is already in use, for system-wide search.  Actually, almost all Windows+letter combinations mean something, so we should avoid checking for it.  Your code will need to check whether it's on mac or windows, before checking for the meta key.

jpolitz:
If no one else gets to it before then, I'll check it at some point over the weekend; I have a Windows machine at home.

YanyanR:
Is it worth using [navigator.platform](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorID/platform) here to detect windows/mac? (It returns a string representing the platform of the browser.) 

Another question, when testing windows vs mac, have you used Safari's User Agent under Develop menu? Does it work well, or it's still better to test on a windows machine?

jpolitz:
I think we should probably look up keybinding libraries if we want to do this right.

I usually want to try on a keyboard that looks like a Windows keyboard for things like this (and also test in Edge).

jpolitz:
Though it's certainly not a _bad_ idea to try out the Safari mode :-) 

We'll know a little more about how good it is if we test in both ways.

YanyanR:
When you say "keybinding libraries", do you mean finding out if the combinations are already in use?
 
If so here's the list for [Windows 10](https://support.microsoft.com/en-us/help/12445/windows-keyboard-shortcuts):
- Windows-S: Open search
- Windows-M: Minimize all windows
- Windows-Enter: Open Narrator
- Windows-?: nothing

But then we'll also need to consider different browsers' shortcut, like [Chrome shortcut](https://support.google.com/chrome/answer/157179?hl=en), and different OS versions (win7 vs win8)? And should we consider linux too?

jpolitz:
I mean like https://craig.is/killing/mice or
https://github.com/RobertWHurst/KeyboardJS

YanyanR:
[This commit](214457f7aea8d8a0e840b958db48039c15ff94d0) is about adding [mousetrap library](https://craig.is/killing/mice), which has a generic `mod` helper that lets you set cross platform shortcuts. On Mac this ends up mapping to `command+s` whereas on Windows and Linux it maps to `ctrl+s`. 

@jpolitz could you test it on a windows machine? Safari's user agent only sets the browser but not the keyboard.

jpolitz:
I'm merging this so I can more easily test it. I started spending way too much time doing the right local install for windows, and I don't think it'll hurt anything for it to be on horizon for testing.

PR #261, opened 2018-02-07
Patch a11y

jpolitz:
Generating PR so we can see a diff and talk about it.


src/web/css/editor-patch.css
```diff
@@ -0,0 +1,1926 @@
...
+  opacity: 1.0;
+  }
+  to {
+  display: block;
+  transform: scale(10);
+  opacity: 0;
+  }
+}
+
+>>>>>>> horizon
```
blerner: Merge conflict?

src/web/css/shared.css
```diff
@@ -14,7 +16,7 @@ em {
   width: 50%;
   height: 100%;
   z-index: 8700;
-  background: url('/img/pyret-spin.gif') 50% 30% no-repeat rgba(249, 249, 249, 0.5);
+  background: url('/img/patch-treadmill-run.gif') 50% 30% no-repeat rgba(249, 249, 249, 0.5);
```
blerner: Does this affect CPO as well as Patch?

src/web/js/google-apis/api-wrapper.js
```diff
@@ -7,14 +7,17 @@
  * The most recently authenticated version of the
  * wrapped Google API
  */
+
```
blerner: These changes all seem commented out...

src/web/js/google-apis/drive.js
```diff
@@ -1,4 +1,5 @@
 window.createProgramCollectionAPI = function createProgramCollectionAPI(collectionName, immediate) {
+  console.log('doing createProgramCollectionAPI (immediate = ' + immediate + ')');
```
blerner: ...and these seem to be entirely debug-prints that should be cleaned up.

src/web/js/output-ui.js
```diff
@@ -926,6 +969,9 @@
             }, "optional: help(contents)");
           },
           "text": function(txt) {
+            if (cpoDialect === "patch") {
```
blerner: I think I would prefer to see a `$cpoPyret` and `$cpoPatch` pair of renderers, rather than a single renderer that's polyglot across the two.  You can have `$cpoPatch` extend from `$cpoPyret` and override just the parts that change...

blerner: (The intent of making a dictionary of these various renderers was that it should be easy to add new ones for new contexts, not shoehorn new features into old renderers.)

src/web/js/trove/world-lib.js
```diff
@@ -650,13 +602,9 @@
...
     };
 
-
-
-
-
     //////////////////////////////////////////////////////////////////////
 
-    var bigBang, StopWhenHandler;
+    var bigBang, StopWhenHandler, LastImageHandler;
```
blerner: Is this the same as or different from your last-image-handler PR?

src/web/js/trove/world.js
```diff
@@ -15,6 +15,8 @@
     values: {
       "reactor": ["forall", ["a"], ["arrow", [["tid", "a"], ["List", "WCOofA"]], "Any"]],
       "big-bang": ["forall", ["a"], ["arrow", [["tid", "a"], ["List", "WCOofA"]], ["tid", "a"]]],
+      "_patch_big-bang": "tany",
```
blerner: Again, really don't like exposing `_pseudoprivate` names.  What are these for?

src/web/js/trove/world.js
```diff
@@ -609,16 +706,40 @@
...
+          }
+          return runtime.makeOpaque(new OnTick(handler, fixN * 1000));
+        }),
+
         "to-draw": makeFunction(function(drawer) {
           runtime.ffi.checkArity(1, arguments, "to-draw", false);
           runtime.checkFunction(drawer);
           return runtime.makeOpaque(new ToDraw(drawer));
         }),
+        "on-redraw": makeFunction(function(drawer) {
```
blerner: What is this handler for?

src/web/css/shared.css
ds26gte: No. shared.css is used only by Pyret (via editor.html). Its sister file, shared-patch.css, is used only by Patch (via editor-patch.html). When running Patch, editor-patch.html is copied to editor.html. For nomenclatural consistency, I've added editor-pyret.html, which is exactly the original editor.html.

A clone of the code.pyret.org repo will run Pyret by default. In order to have it run Patch, copy editor-patch.html to editor.html, and have the pyret-mode.js line in the Makefile point to patch.js rather than pyret.js. This is two changes, and it would of course be nice if it could be reduced into just one change, presumably just the Makefile where the 'cp editor-patch.html editor.html' can be inserted. My Makefile skills are a bit rusty: Maybe a dialect argument to the Makefile can avoid the need for having to modify the Makefile at all.

src/web/js/cpo-main.js
```diff
@@ -267,19 +299,30 @@
...
               runtime.runThunk(function() {
-                return runtime.safeCall(
-                  function() {
-                    return gf(repl,
-                    "make-interaction-locator").app(
-                      runtime.makeFunction(function() { return str; }))
-                  },
-                  function(locator) {
+                //console.log('doing run timeout');
+                if (cpoDialect!=='patch') {
```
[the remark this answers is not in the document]
ds26gte: Sure. If using a switch statement, would you prefer Pyret to simply be the default: case, or would you like a specific label ('pyret') for it?  (My preference is that Pyret simply be the default, since it's presumably always going to be the "default" dialect, in that it won't require any translation layer. It's a purely aesthetic choice to be sure.)

src/web/js/trove/image.js
```diff
@@ -368,11 +390,70 @@
...
+      var c = checkColor(maybeColor);
+      return runtime.wrap(image.colorBlue(c));
+    });
+    f("color-alpha", function(maybeColor) {
+      checkArity(1, arguments, "color-alpha");
+      var c = checkColor(maybeColor);
+      return runtime.wrap(image.colorAlpha(c));
+    });
+
+    f("bitmap-url", bitmapURL);
```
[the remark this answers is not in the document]
schanzer: I suspect @ds26gte added it for backwards compatibility with BS:A. However, it certainly shouldn't be in the image library itself -- Racket calls to `bitmap/url` should be translated to `image-url`.

src/web/js/trove/world.js
schanzer: This one is almost certainly for compatibility with really old WeScheme programs. The design goal of Patch is that any WeScheme program that ran before should continue to run, and some Very Old WeScheme programs used `on-redraw` instead of `to-draw`

src/web/css/shared.css
ds26gte: I see what you mean. The changes to shared.css in the patch-a11y branch should not be merged. (They are not used by Patch anyway for a long time, all the Patch-specific changes having been consolidated into the new file shared-patch.css, but I neglected to undo my changes to shared.css.)

src/web/js/trove/world-lib.js
ds26gte: Same.

src/web/js/trove/world.js
ds26gte: Yeah, Patch's big-bang procedure is not quite the same as Pyret's, although it uses the same name. There are quite a few such procedures where Patch prefers a slightly different signature, as you've noticed.

I used _patch_big-bang as the name of the Pyret procedure that Patch's big-bang would map to. It is a bona-fide procedures and _could_ be used in Pyret (i.e., they won't cause errors if used with appropriate arguments, so it's not really private). I can update it and similar procedures to use your new-image-lib eventually so as not to clutter the Pyret namespace. The naming convention (they all start with _patch_).

src/web/js/cpo-main.js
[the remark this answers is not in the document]
blerner: I might probably write it as 
```js
switch(cpoDialect) { 
case "Patch": 
  ...patch specific stuff...;
  break;
case "Pyret":
default:
  ...Pyret specific stuff
}
```
and get the best of both worlds :)

src/web/css/shared.css
blerner: Right.  This is the `shared.css` file, not the `shared-patch.css` file.

src/web/js/trove/world.js
blerner: This seems like something perhaps a Patch-specific wrapper around world.js might achieve, rather than baking it into the underlying world.js library that's exposed to Pyret?

blerner:
Closing -- the Patch components should be pulled to their own PR if necessary, and the a11y components have been merged in #268.

PR #268, opened 2018-04-12
added accessibility code

ds26gte:
This addresses ds26gte/code.pyret.org#87. Key features:

- Screenreader extensions trigger appropriate audio information as a blind user navigates the screen.
- With the use of aria-labels, four broad regions are identified in the editor window (Toolbar, Definitions, Interactions, and Announcements). Use F6 and Shift-F6 to cycle focus amongst them.
- Alt-n to review items in interaction history.
- Left and right arrow read specific character at cursor.

### DIFF
--- a/src/web/js/repl-ui.js
+++ b/src/web/js/repl-ui.js
@@ -85,7 +85,7 @@
 
     // the result of applying `displayResult` is a function that MUST
     // NOT BE CALLED ON THE PYRET STACK.
-    function displayResult(output, callingRuntime, resultRuntime, isMain) {
+    function displayResult(output, callingRuntime, resultRuntime, isMain, updateItems) {
       var runtime = callingRuntime;
       var rr = resultRuntime;
 
@@ -104,6 +104,9 @@
                 html.addClass("highlights-active");
               });
               html.addClass('compile-error').appendTo(output);
+              if (updateItems) {
+                updateItems();
+              }
               if (click) html.click();
             }).done(function () {restarter.resume(runtime.nothing)});
         });
@@ -127,6 +130,9 @@
           }
           else if(callingRuntime.isSuccessResult(result)) {
             result = result.result;
+            if (updateItems) {
+              updateItems();
+            }
             return ffi.cases(ffi.isEither, "is-Either", result, {
               left: function(compileResultErrors) {
                 closeAnimationIfOpen();
@@ -184,6 +190,9 @@
             });
           }
           else {
+            if (updateItems) {
+              updateItems();
+            }
             doneDisplay.reject("Error displaying output");
             console.error("Bad result: ", result);
             didError = true;
@@ -216,10 +225,10 @@
       var pointer = -1;
       var current = "";
       function loadItem() {
-        CM.setValue(items[pointer]);
+        CM.setValue(items[pointer].code);
       }
-      function saveItem() {
-        items.unshift(CM.getValue());
+      function saveItem() { //not used?
+        items.unshift({code: CM.getValue(), output: false});
       }
       function prevItem() {
         if (pointer === -1) {
@@ -242,6 +251,48 @@
           pointer--;
         }
       }
+
+      // a11y stuff
+
+      function say(msg, forget) {
+        if (msg === "") return;
+        var announcements = document.getElementById("announcementlist");
+        var li = document.createElement("LI");
+        li.appendChild(document.createTextNode(msg));
+        announcements.insertBefore(li, announcements.firstChild);
+        if (forget) {
+          setTimeout(function() {
+            announcements.removeChild(li);
+          }, 1000);
+        }
+      }
+
+      function sayAndForget(msg) {
+        //console.log('doing sayAndForget', msg);
+        say(msg, true);
+      }
+
+      function speakHistory(n) {
+        if (n === 0) { n = 10; }
+        var historySize = items.length;
+        console.log('historySize =', historySize);
+        if (n > historySize) { return false; }
+        var history = items[n-1];
+        sayAndForget(history.code + (history.output ?
+          " evaluates to " + history.output :
+          " resulted in an error." + history.erroroutput));
+        return true;
+      }
+
+      function speakChar(cm) {
+        var pos = cm.getCursor();
+        var ln = pos.line; var ch = pos.ch;
+        var char = cm.getRange({line: ln, ch: ch}, {line: ln, ch: ch+1});
+        sayAndForget(char);
+      }
+
+
+      // end a11y stuff
 
       container.append(mkWarningUpper());
       container.append(mkWarningLower());
@@ -476,6 +527,26 @@
         "vertical-align": "middle"
       });
       var runContents;
+      function updateItems() {
+        var thiscode = items[0];
+        var docOutput = document.getElementById("output");
+        var lastOutput = docOutput.lastElementChild;
+        var text;
+        if (lastOutput.classList.contains('compile-error')) {
+          var pList = lastOutput.getElementsByTagName('p');
+          text = '';
+          for (var i = 0; i < pList.length; i++) {
+            text += '. ' + pList[i].innerText;
+          }
+          thiscode.erroroutput = text;
+          thiscode.output = false;
+          sayAndForget(thiscode.code + ' resulted in an error. ' + text);
+        } else {
+          text = lastOutput.innerText;
+          thiscode.output = text;
+          sayAndForget(thiscode.code + ' evaluates to ' + text);
+        }
+      }
       function afterRun(cm) {
         return function() {
           running = false;
@@ -671,7 +742,8 @@
       var runner = function(code) {
         if(running) { return; }
         running = true;
-        items.unshift(code);
+        var thiscode = {code: code, output: false, erroroutput: false};
+        items.unshift(thiscode);
         pointer = -1;
         var echoContainer = $("<div class='echo-container'>");
         var echoSpan = $("<span>").addClass("repl-echo");
@@ -694,7 +766,7 @@
           maybeShowOutputPending();
           return r;
         });
-        var doneRendering = startRendering.then(displayResult(output, runtime, repl.runtime, false)).fail(function(err) {
+        var doneRendering = startRendering.then(displayResult(output, runtime, repl.runtime, false, updateItems)).fail(function(err) {
           console.error("Error displaying result: ", err);
         });
         doneRendering.fin(afterRun(CM));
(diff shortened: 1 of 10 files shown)


schanzer:
A few issues (tested on VoiceOver on OSX):
- the Pyret dropdown menu and the Run options menu describe themselves as "small downwards pointing triangle". Both need a proper ARIA-label
- while the spacebar toggles the display of the dropdowns, there is currently no way to aurally-navigate the submenus themselves 
- When navigating the toolbar, the run options dropdown comes _before_ run. Run should come first
- Output from clicking "Run" is not read aloud - neither evaluated output or error messages
- Test reporting is not read aloud

src/web/js/repl-ui.js
```diff
@@ -85,7 +85,7 @@
 
     // the result of applying `displayResult` is a function that MUST
     // NOT BE CALLED ON THE PYRET STACK.
-    function displayResult(output, callingRuntime, resultRuntime, isMain) {
+    function displayResult(output, callingRuntime, resultRuntime, isMain, updateItems) {
```
jpolitz: I know this is a bit hypocritical to ask here, given the lack of comments on the _other_ parameters, but could we get a brief description/comment on what `updateItems` is for?

src/web/js/repl-ui.js
```diff
@@ -216,10 +225,10 @@
       var pointer = -1;
       var current = "";
       function loadItem() {
-        CM.setValue(items[pointer]);
+        CM.setValue(items[pointer].code);
       }
-      function saveItem() {
-        items.unshift(CM.getValue());
+      function saveItem() { //not used?
```
jpolitz: Yup, agreed, I don't see any use of this.

jpolitz:
Thanks for splitting this off. @blerner I don't see anything radical here that should prevent a merge; this seems essentially orthogonal and not likely to cause any issues.

src/web/js/repl-ui.js
ds26gte: Will do. BTW, this PR will be gaining more updates. Do you want to wait to merge, or is it best to merge early (and often)?

jpolitz:
Actually, could I just get a brief "script" of what I should do, manually, to make sure this is working? Even having that bulleted list would go a long way.

ds26gte:
It's intended as an aural experience that's cleanly optional. Bring up the Pyret IDE in a browser (e.g., Chrome), turn on a screenreading extension (e.g., ChromeVox), and confirm that it reports aurally reasonable things as you interact, i.e, what the input expression is and what it evaluated to, enriching the latter with type information instead of just reading text. 

E.g., "(list true 3.14) evaluates to a list of two items: true, a boolean, and 3.14, a rational number".

In addition, F6 and shift-F6 should cycle through the major divisions on the window (again letting you know aurally where you are, if the screenreader is on), and Tab and Shift-Tab should cycle through the items within a division where it makes sense. In the interaction window, Alt-<num> should recount what happened <num> interactions back. 

You can also micro-edit the current input expression using the arrow keys, in which case it should tell you what character is being cursored over. (This makes editing easier for blind users.)

jpolitz:
Things I noticed in Chrome/OSX (sorry if these are known issues; if there's a summary of working/not working that would be great).

- I type in 

```
x + 1
```

into interactions, with no definition of x. I hear the error message read aloud to me (cool) though it skips the snippet (oh well). Then it says "x + 1 evaluates to x + 1". This last part seems wrong.

- Values that are printed from definitions as top-level values aren't dictated.

- It seems like the `ariaText` field is present on images (awesome!), but this PR doesn't put that information in the say-able output.

schanzer:
@jpolitz one way to get some automated testing is to check the `aria-label` attribute of DOM nodes rendered in the REPL. For testing F6 cycling, you can compare `document.activeElement` against expected values. These don't get you all the way, but they get pretty far.

I've added your error reports as issues on @ds26gte 's fork. He and I have traded mail about some of them, but now that you and Ben are looking at them it's probably time to track them in public. 

Runtime values will eventually declare their `ariaText` fields, but that won't happen until the corresponding PR for `pyret-lang`.

cse131-s18:
All sounds good

> Runtime values will eventually declare their ariaText fields, but that won't happen until the corresponding PR for pyret-lang.

Got it, a pointer to the core of those changes would be helpful

I'd like to know why a change to pyret-lang is necessary here; it seems like this should be handle-able at the rendering level rather than at the value level. That is, we should be able to write a function that takes in a ValueSkeleton and produces a "say" of the text within it, with no changes to how values in the language are represented. If not, we should make sure ValueSkeleton _can_ do that, because it's where we try to have all user-visible value stuff live.

schanzer:
Still using the class account, eh? :)

Can you say a little more about what kind of information a valueskelton contains? For example, is the entire scene graph for an image preserved inside of it? If so, this seems like a reasonable plan.

jpolitz:
> Still using the class account, eh? :)

I _will_ master browser profiles some day.

The skeleton contains a terminal node called "vs-value", which contains any Pyret value. The renderer walks the skeleton, and when it reaches values, can decide to handle them specially or try to get a skeleton for them recursively. In the case of images, CPO checks specifically for image-lib image values and does the right thing to generate the canvas for them; this has the reference to the image and can traverse its internal image structure, or just grab the ariaText off of the image value.

schanzer:
Ah, excellent! Then yeah, this should be doable entirely within CPO, through ValueSkeletons.

jpolitz:
Awesome additions, spent a little more time with this tonight. In no particular order, and with a bit of admitted obliviousness about what "should work":


I notice that when I use the up/down keys to cycle through interaction history, it doesn't say what's happening. I don't quite get this comment (if it's my user error or misunderstanding):

> In the interaction window, Alt- should recount what happened interactions back.

I tried alt+1, alt+2 and nothing seemed to happen, but I'm not sure I should expect it to.



Images worked great.



The ValueSkeleton constructors work great! I was pleasantly surprised when I wrote:

```
> import string-dict as SD
> s = [SD.string-dict: "a", 1, "b", 7]
> s
```

And it said "s evaluates to a string hyphen dict of four items a, a string, one, b, a string, seven" (the four items bit isn't the a11y's fault, that's just a weakness of string-dicts).

That's awesome! Having the name of the constructor separate from the contents is terrific, and works nicely generically across collections.

Unfortunately, doing it for the tuply string-dict library didn't work:

```
> import string-dict as SD
> s = [SD.string-dict: {"a"; 1}, {"b"; 3}]
> s
```

This says "undefined" for the elements, and I'm not sure why. Tuples and tables don't really work yet, which is probably the reason. Still, this gives me lots of happy feelings about the way value rendering works as a datatype with this – we didn't have to add _anything_ to the string-dict library to make this work.



The errors now mention the snippet code (awesome), but still repeat the text back at the end as "y + 6 evaluates to y + 6".

Errors from definitions aren't spoken that I can tell.


This is a problem perhaps out of scope for us for now, but syntax errors are just a nightmare! If I type in `(x + 1` (note the missing paren, sorry SK), the error says something like

"Pyret didn't expect your program to end as soon as it did in _ex plus one_"

with no mention of the actual punctuation that caused the issue. Maybe we need it to go into character-by-character mode for the snippets (this is the worst aural version of the problem students have when not "seeing" the missing colon!). I assume someone else has thought hard about this problem, so hopefully there's a best practice we can draw from.


I cannot always reproduce this, but I've repeatedly managed to get it "stuck" in a state where it didn't say the interactions. I'm pretty sure it has to do with window focus, so it may be an OS level issue. But if I tab away, edit this comment that I'm writing, then go back, interactions doesn't say things back to me anymore until I refresh the page. I feel like it happened without switching once, but I'm not sure. It definitely happens most of the time when I switch to another tab and come back – not sure if there's a refocus API or similar that would make this work.

jpolitz:
More generically, @schanzer , regarding your comment about needing image structure to do this well, I think that it could well be the right way to approach this from a language design perspective is that ValueSkeleton gets enriched in ways we need to do aural output. Its goal is to represent the user-facing side of values (which is already working well with constructors!), so it might be a good idea to put optional fields on VS that help aria speak. @ds26gte already found some good defaults for the collection bits, but there is opportunity for user control as well.

One way to look at this is that each vs-value could come with a string description that can be vocalized. This could be useful as alt-text on a rendering, an alternative to images in text-based environments, and also be included as aria text. Then it would be an image's job to render as a vs-value with the appropriate string argument.

Now, images are pretty primitive anyway in CPO, but it's a reasonable way to generalize this, and might be part of the way forward for making tables and rows work, where we won't just want to start dictating all the values necessarily.

schanzer:
I see that CPO already inherited WeScheme’s ariaText code for the image library - so right now that logic doesn’t live in the valueskeleton. Do you see this as something that needs to be moved?

schanzer:
Clicking "Run" when the following code is in the definitions window reads 'true', but not the star.

```
include image
star(50, "solid", "red")
true
```

This happens regardless of whether `true` comes before or after the star.

I'd say that once definitions are read reliably and the UI is fully navigable, we're at a local maximum where it would make sense to merge.

ds26gte:
@jpolitz, there does seem to be some browser- or extension-specific dependence.  Erroring expressions are not supposed to say they evaluate to themselves -- and they never did for me, even before the latest addition. Could you state your brownser/extension combo? (I'm using Ubuntu/Chrome/ChromeVox.) The interaction history keeps track of three things for each interaction: code, output, and erroroutput. It looks like something is setting output rather than erroroutput because the error in the output is being recognized later than expected, but even so I can't see why the output is considered to be the input expression. (A somewhat similar thing happens, by intent, with (just) the error snippet: it is set very late in the game, so it's not available when the history is updated. I punted on that by simply reusing the input expression whenever an error occurs, but this is embedded inside the erroroutput with surrounding error verbiage, and would not cause the problem you're seeing.)

Also: Alt-num is one of the key things that should work. Alt-3 should report what happened 3 interactions back, error or non-error, exactly as though you typed that same input now.

ds26gte:
@schanzer, I wasn't voicing the error snippet before (the surrounding error message is plausibly sufficient information to a blind user, and a case could me made that the error snippet is redundant, since it seems to repeat the input expression in toto).  I actually had to do some special effort to voice it because @jpolitz said he could feel its lack in the recitation. Dropping it would be quite easy if there's consensus on that.

jpolitz:
> I see that CPO already inherited WeScheme’s ariaText code for the image library - so right now that logic doesn’t live in the valueskeleton. Do you see this as something that needs to be moved?

Not urgently. It's a nice design opportunity in the future.

> I actually had to do some special effort to voice it because @jpolitz said he could feel its lack in the recitation.

I think _something_ is necessary here, and the snippet is way better than the nothing from before, IMO. It's currently _way_ improved for, e.g. arity mismatch errors:

"This application expression errored.  In string hyphen equals a 3. Two arguments were passed to the operator.  The operator evaluated to a function accepting one parameter."

The snippet at least tells you what expression was involved in the error, it's just that it doesn't read it "as code". I wonder if reading the snippet and then voicing the location (e.g. "ex was used but not defined in three times ex plus one in definitions at line four column six to line four column twelve") would be useful.

jpolitz:
@ds26gte I'm using Chrome 66.0.3359.117 and the stock VoiceOver application on OSX 10.13.4

schanzer:
@jpolitz you want to use ChromeVox Next. Regular chromeVox is many years old, and pretty much on supported

blerner:
I just tried pulling this onto my local horizon, and full-rebuilding.  Several major problems:
* None of the menus dismiss when clicked -- not the Bonnie menu, or the Run options menu.  (They used to disappear by clicking on them again.)
* The Run menu is way too narrow to read, and a different color than the Run button
* Mousing over the menus in the toolbar adds a border to them, which causes them to jiggle right by a pixel or two.
* Clicking Run when the definitions area is empty will hang the page.
image
The console is complaining about lastOutput not is null (in Firefox) or classList is null (in Chrome).  I'm guessing this JS error is what's causing the page to hang.
* Refresh the page.  Enter `1 + 2` into the definitions area, and click Run.  (Get 3 as the output.)  Press the up arrow in the REPL, and get `def//` -- what is that??

jpolitz:
You can run selenium tests locally with

```
$ heroku local:run ./node_modules/mocha/bin/mocha
```

You may need to configure the `GOOGLE_CHROME_BINARY` environment variable to point at a path on your system that has Google Chrome (any modern version is fine) installed. The selenium packages should be installed through `npm`.

We're seeing test failures on the `type-check` section of tests because of the empty program issue quoted above.

jpolitz:
Closing this as it was merged through the `merge-a11y` branch as a single (revertable-if-necessary) commit.

PR #272, opened 2018-07-03
New image library

blerner:


src/web/js/output-ui.js
```diff
@@ -178,11 +178,11 @@
       Position.fromPyretSrcloc = function (runtime, srcloc, loc, documents, options) {
         return runtime.ffi.cases(isSrcloc, "Srcloc", loc, {
           "builtin": function(_) {
-             throw new Error("Cannot get Position from builtin location", loc);
+             return new Error("Cannot get Position from builtin location", loc);
```
jpolitz: These seem like weird changes to have in the images update.

package.json
```diff
@@ -43,6 +43,7 @@
     "karma-sourcemap-loader": "0.3.7",
     "karma-webpack": "^3.0.5",
     "load-script": "^1.0.0",
+    "lodash": "~2.4.1",
```
jpolitz: Where is lodash used?

src/web/js/output-ui.js
blerner: I mostly agree.  However, IIRC the commit that introduced that (https://github.com/brownplt/code.pyret.org/commit/1eced4f62b1c15e35b20bf93b3d8893b4231a80f) was trying to recover from some weird stack issue that triggered in deeply-nested images when rendered in error messages, or something like that, and it boiled down to fixing the stack-unsafety of this case.  This change should be cherry-picked anyway, regardless of this PR.

blerner:
No idea. I don't even remember editing package.json...

jpolitz:
- [ ] Make sure type and untyped provide the same functions
- [ ] Check naming of new files for import/include back-compat
- [ ] Check on needing to import image-structs

PR #283, opened 2018-08-06
more a11y enhancements

ds26gte:
Replaced float mechanisms used in the nav bar with CSS3 flex box. Reasons:

1. Floats require the HTML (i.e., DOM) to invert the flushright menuitems, which have to be carefully re-un-inverted in JS when coding user navigation and screen-reading. Flex box removes this complexity.

2. Maintainable.

Also added some more help text for C-?.  C-? is now advertised in the Bonnie submenu.

### DIFF
--- a/src/web/css/editor.css
+++ b/src/web/css/editor.css
@@ -1264,7 +1264,15 @@ code div {
 #runPart {
   background: #317BCF;
   padding: 0;
-  margin: 0;
+  /* margin != 0 because we'll be pushing this flex item right */
+}
+
+#runli {
+  z-index: 2;
+}
+
+#rundropdownli {
+  z-index: 1; /* to push its left lap under the Run button */
 }
 
 #rundropdownli {
--- a/src/web/css/editor.css
+++ b/src/web/css/editor.css
@@ -117,6 +117,7 @@ div.CodeMirror span.CodeMirror-nonmatchingbracket-region {
 
 #REPL {
   font-family: sans-serif;
+  background: #f8f8f8;
 }
 
 .replTextOutput {
@@ -506,6 +507,7 @@ body {
   box-sizing: border-box;
   -moz-box-sizing: border-box;
   margin:0;
+  background: #f8f8f8;
 }
 
 .replContainer {
--- a/src/web/css/editor.css
+++ b/src/web/css/editor.css
@@ -1261,6 +1261,14 @@ code div {
   box-shadow: none;
 }
 
+#breakButton:disabled {
+  background: #C8C8C8;
+}
+
+#breakButton:hover[disabled] {
+  background: #C8C8C8;
+}
+
 #runPart {
   background: #317BCF;
   padding: 0;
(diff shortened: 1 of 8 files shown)


blerner:
Can you confirm that the use of `display: flex` doesn't affect the minimum browser versions we support?  (Based on https://developer.mozilla.org/en-US/docs/Web/CSS/flex, I think we're good...)  It worries me a bit that you haven't specified `flex-direction` anywhere, so there's a bit of implicit behavior here, but that may be ok...

ds26gte:
Is there a list of browser requirements somewhere? I did a `grep -r` for sundry browser names in the repos, but a standard list would be better of course. In any case, code.pyret.org already uses Flexbox in the check-error rendering. So we should reassuringly fail on several fronts if Flexbox isn't supported. :-)

`flex-direction: row` (= what we want) is _guaranteed_ to be the default if Flexbox is supported at all. But I can add it if you like. (I think it's better future-proofing to trust a system default than to hard-wire a setting in case Pyret gets ported to Hebrew or traditional Japanese.)

I'll also experiment with a cell phone to see if `flex-wrap: wrap` makes sense. It shouldn't hurt for large screens but I should check.

ds26gte:
I've added `flex-flow: row wrap` to the Toolbar flexbox.  (`row` is default, `wrap` isn't, but I haven't found the Toolbar to not wrap, regardless of this setting. Browsers checked: Chrome, Firefox, Opera. But explicit wrapping doesn't hurt.)

Please note that wrapping the Toolbar does occlude the top of the Def and Interaction windows. This can't be helped because the Toolbar is like a permanent awning from the browser top edge and doesn't really "push" the Def and Int windows down. (Perhaps some JS could be added to dynamically or occasionally check the height of the Toolbar and correspondingly change the absolute positioning of the Def/Int windows, but the payoff is paltry.)

Cellphone test doesn't help much: The browser window is not resizable (it feels more like a PDF file that can be panned), so an opportunity to wrap doesn't arise, at least on my phone. In any case, wrapping would have been an issue only for cellphones and small tablets, but I don't have enough test devices to actually create the issue. If anyone runs into it, be sure to let me know of course.

ds26gte:
Squishing the Pyret window horizontally so that the Toolbar wraps to a second line (or, in principle, more) will now cause the Definitions and Interactions windows to be pushed down correspondingly, so everything in those windows stays visible. (The color of the leftover area in the spillover line(s) matches the Def/Int windows to avoid jarring the user.)

This is not a problem thrown up by this PR.  We've always had this issue.

ds26gte:
5da4f6a fixes brownplt/pyret-lang#416.

`Up/Dn` arrows traverse REPL history to populate the current prompt with an old input. These now skip all but one of contiguous duplicate inputs. However, `Alt-n` which traverses REPL history in order to recite it for the benefit of a blind user, should not skip duplicates.

When populating the current prompt, the re-used input is said aloud for the benefit of a blind user.

jpolitz:
Going ahead and merging this to review by looking at horizon directly. Thanks for updates!

PR #285, opened 2018-08-08
Fix 1350

jpolitz:


jpolitz:
Yay tests! These aren't running b/c this is pointing at horizon which doesn't have the updates yet. We can re-trigger the build once https://github.com/brownplt/pyret-lang/pull/1386 lands to have an automated check.

PR #303, opened 2019-02-11
Josh/nolan localfilesave

j3ven7:
First merge


schanzer:
@blerner I'm assuming that

1. CPO has changed enough in the last half-decade that this PR is no longer usable, and
2. That with Anchor coming down the pike, this will also soon be moot.

OK to close?

blerner:
I've never seen this code before. I don't know what it is intended to do...

jpolitz:
This was a great idea at the time. It will be superseded by other save-to-localstorage/indexedDB solutions before we can make use of it, though.

jpolitz:
(This was actually really great work that went a little beyond just local file saving. It was an electron-hosted container for CPO for offline use.)

blerner:
Well that does sound cool! 👍

jpolitz:
Actually now I'm realizing it still has some good ideas. Looking at the diff is probably useful, and actually only package* conflict. It's kind of orthogonal to what's going on on the page. And since we have BrowserFS presumably we can swap in real `node` implementations of that API. Not a totally outlandish experiment to try and finish.

jpolitz:
I think this PR is now all subsumed by the VScode extension (https://marketplace.visualstudio.com/items?itemName=PyretProgrammingLanguage.pyret-parley), which borrowed a lot of ideas from this work.

Thanks @dlhsiao / @nolangray15 / Josh!

PR #343, opened 2020-09-07
horizon -> master, themes and other small fixes

jpolitz:


jpolitz:
Oh, the other big update here is the change to how we use the gdrive library on the server when proxying is necessary to avoid Google blocking us as a bot for making unauthenticated requests rapidly. Noting that for posterity.

PR #347, opened 2020-09-16
Skip blanks at the REPL

jpolitz:
This marks *all* blanks at the REPL as skips, and generalizes duplicates to skips

This commit also fixes a bug with advanceDups and nextItem, which would never recall the current uncommitted repl entry if the repl history ended in a sequence of skips

https://github.com/brownplt/pyret-lang/issues/1529

### DIFF
--- a/src/web/js/repl-history.js
+++ b/src/web/js/repl-history.js
@@ -15,7 +15,7 @@
       var thisItem;
       while (this.pointer >= 0 && this.pointer < this.items.length) {
         thisItem = this.items[this.pointer];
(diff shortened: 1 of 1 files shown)


jpolitz:
Added @blerner and @ds26gte as reviewers (some of this code that I'm re-organizing I believe was written by @ds26gte , but no worries if that's not true enough anymore to review).

jpolitz:
This got merged outside of this PR (@blerner and I looked at it today), so closing.

PR #355, opened 2020-11-28
Add minor gridlines

sorawee:
Per @schanzer's request.

For example, for the program:

```
include chart
import color as C

fun some-fun(x): num-sin(2 * x) end
a-series = from-list.function-plot(some-fun)
  .color(C.purple)
a-chart-window = render-chart(a-series)

a-chart-window
  .x-axis("this is x-axis")
  .y-axis("this is y-axis")
  .display()
```

In CPO, it displays:

Screen Shot 2020-11-28 at 14 25 43

With this patch, it displays:

Screen Shot 2020-11-28 at 14 27 01

Here's another example for the 10x + 7 graph:

Screen Shot 2020-11-28 at 14 32 08

vs 

Screen Shot 2020-11-28 at 14 35 02

Notice that the number of lines in the minor grids are not the same (across chart / across axes). It might be possible to do that by using a right Google Charts configuration, but that's future work.

### DIFF
--- a/src/web/arr/trove/chart.arr
+++ b/src/web/arr/trove/chart.arr
@@ -94,6 +94,10 @@ legend-method = method(self, legend :: String):
   self.constr()(self.obj.{legend: legend})
 end
 
+show-minor-grid-lines-method = method(self, is-showing :: Boolean):
+  self.constr()(self.obj.{show-minor-grid-lines: is-showing})
+end
+
 x-axis-method = method(self, x-axis :: String):
   self.constr()(self.obj.{x-axis: x-axis})
 end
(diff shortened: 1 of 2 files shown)


blerner:
I'd be happier if the minor-gridlines were a builder method that we could toggle on and off, rather than something permanently on.  E.g. something like `a-chart-window.show-minor-gridlines(true/false)`.  These gridlines look very busy to me, and I'd imagine several folks would want them configurably off.  (I don't mind if the default is on, or if @schanzer can easily tweak the background code to turn them on, assuming that students won't write the chart-window code themselves.)

sorawee:
I can make the change tonight, unless anyone takes over the PR.

jpolitz:
I'm going to merge this because it looks good, addresses Ben's feedback, and I expect it to pass on travis-ci.com after the merge. Thanks @sorawee !

jpolitz:
Up now on code.pyret.org

PR #360, opened 2020-12-23
Add boxplot range

jpolitz:


jpolitz:
@schanzer this LGTM. One thought is that teachpacks that wrap it may want to do `max <= min` checking; the current behavior is that they seem to both work-ish

image

schanzer:
Good call! Will do.

schanzer:
@jpolitz how does one throw exceptions from the JS world? Or would this checking be done in `chart.arr`? It seems like there's no opportunity to do the checking in Pyret-land, but I could be wrong about that.

jpolitz:
Wow sorry I missed this comment.

JS actually has access to make the best error messages. For something that's a function/method argument, you can emulate what images do. `checkArgsInternalXXX` is the best series of functions.

https://github.com/brownplt/code.pyret.org/blob/horizon/src/web/js/trove/make-image.js#L105
https://github.com/brownplt/code.pyret.org/blob/horizon/src/web/js/trove/internal-image-untyped.js#L247

jpolitz:
`checkArrayIndex` is another similar function in the runtime:

https://github.com/brownplt/pyret-lang/blob/horizon/src/js/base/runtime.js#L4039

PR #367, opened 2021-01-13
Add boxplot range

schanzer:
Add type annotations for boxplot min/max methods

### DIFF
--- a/src/web/arr/trove/chart.arr
+++ b/src/web/arr/trove/chart.arr
@@ -406,10 +406,10 @@ data DataSeries:
     method show-outliers(self, show):
       self.constr()(self.obj.{show-outliers: show})
     end,
-    method min(self, min :: Number): -> Option<Number>
+    method min(self :: DataSeries, min :: Number): -> Option<Number>
       self.constr()(self.obj.{min: some(min)})
     end,
-    method max(self, max :: Number): -> Option<Number>
+    method max(self :: DataSeries, max :: Number): -> Option<Number>
       self.constr()(self.obj.{max: some(max)})
     end
   | histogram-series(obj :: HistogramSeries) with:
--- a/src/web/arr/trove/chart.arr
+++ b/src/web/arr/trove/chart.arr
@@ -406,10 +406,10 @@ data DataSeries:
     method show-outliers(self, show):
       self.constr()(self.obj.{show-outliers: show})
     end,
-    method min(self :: DataSeries, min :: Number) -> Option<Number>:
+    method min(self, min :: Number): -> Option<Number>
       self.constr()(self.obj.{min: some(min)})
     end,
-    method max(self :: DataSeries, max :: Number) -> Option<Number>:
+    method max(self, max :: Number): -> Option<Number>
       self.constr()(self.obj.{max: some(max)})
     end
   | histogram-series(obj :: HistogramSeries) with:
--- a/src/web/arr/trove/chart.arr
+++ b/src/web/arr/trove/chart.arr
@@ -406,10 +406,10 @@ data DataSeries:
     method show-outliers(self, show):
       self.constr()(self.obj.{show-outliers: show})
     end,
-    method min(self :: DataSeries, min :: Number): -> Option<Number>
+    method min(self :: DataSeries, min :: Number) -> Option<Number>:
       self.constr()(self.obj.{min: some(min)})
     end,
-    method max(self :: DataSeries, max :: Number): -> Option<Number>
+    method max(self :: DataSeries, max :: Number) -> Option<Number>:
       self.constr()(self.obj.{max: some(max)})
     end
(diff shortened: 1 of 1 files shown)


src/web/arr/trove/chart.arr
```diff
@@ -398,10 +406,10 @@ data DataSeries:
     method show-outliers(self, show):
       self.constr()(self.obj.{show-outliers: show})
     end,
-    method min(self, min):
+    method min(self :: DataSeries, min :: Number) -> Option<Number>:
```
sorawee, line 409: IIUC (it's been long time I last touched this code), the return type should be a `DataSeries`. But you can also leave it out, since most other methods also return a `DataSeries` and they didn't have the return type annotation.

I'm surprised at the conversation in Slack and this PR. Why don't you get a syntax error when you have a syntactically invalid program? And why doesn't this PR error to you? Are you sure that you have the right environment for testing (say, not testing a wrong branch or something like that)?

src/web/arr/trove/chart.arr
```diff
@@ -443,6 +451,8 @@ data ChartWindow:
     constr: {(): box-plot-chart-window},
     x-axis: x-axis-method,
     y-axis: y-axis-method,
+    min: min-method,
```
sorawee, line 454: This will not work, because `default-box-plot-chart-window-object` doesn't have `max` and `min` fields.

More generally, are `max` and `min` properties of the chart window, or properties of data?

schanzer: They're the settings of the chartwindow - but I thought any settings there had to apply to every chart type?

sorawee: Oh, ok. Sorry about "an error will be thrown". That's not true. It's true however that this will not work as intended.

sorawee: In that case, https://github.com/brownplt/code.pyret.org/commit/a00cefc5bf46a8e0660d887504a9b7e98e460415 should be reverted, and

```
  min :: Option<Number>,
  max :: Option<Number>
```

and 

```
  min: none,
  max: none
```

should be added to `BoxChartWindowObject` and `default-box-plot-chart-window-object` respectively. The JS code is also needed to be modified to retrieve the information from `globalOptions` instead of `rawData`.

As for this PR, 

```
    method max(self :: DataSeries, max :: Number) -> Option<Number>:
      self.constr()(self.obj.{max: some(max)}
    end
```

should be removed entirely. Adding `min: min-method,` is the correct way to do it.

schanzer: @sorawee I'm with you up until the last part. Are you saying I should rewrite:

```
    method min(self :: DataSeries, min :: Number):
      self.constr()(self.obj.{min: some(min)})
    end,
    method max(self :: DataSeries, max :: Number):
      self.constr()(self.obj.{max: some(max)})
    end
```

as

```
min: min-method,
max: max-method
```
?

sorawee: Not really. I meant:

```
    method min(self :: DataSeries, min :: Number):
      self.constr()(self.obj.{min: some(min)})
    end,
    method max(self :: DataSeries, max :: Number):
      self.constr()(self.obj.{max: some(max)})
    end
```

should be removed entirely, since you want `min` and `max` to operate on `ChartWindow`, not `DataSeries`. And

```
min: min-method,
max: max-method
```

which operates on `ChartWindow` in the current PR is correct already.

src/web/arr/trove/chart.arr
```diff
@@ -398,12 +406,8 @@ data DataSeries:
     method show-outliers(self, show):
       self.constr()(self.obj.{show-outliers: show})
     end,
-    method min(self, min):
-      self.constr()(self.obj.{min: some(min)})
-    end,
-    method max(self, max):
-      self.constr()(self.obj.{max: some(max)})
-    end
+    min: min-method,
```
sorawee, line 409: @schanzer @jpolitz As discussed earlier, this line (along with the next line) should be removed. The JS side also needs to be updated to extract information from the chart window object rather than data series.

sorawee: I submitted #371 to clarify my proposal. Please review and discuss.

jpolitz:
@sorawee @schanzer I thought this was done, but maybe I didn't understand. Can you clarify if this is _broken_ right now, or of you're suggesting a refactoring?

sorawee:
It's not broken in a sense that if people want to use it, they can make it work. It's broken in a sense that it's not the API that we want to provide to users, and if people start using it, it might be difficult to fix later.

sorawee:
Well, one thing that is functionally broken is that people can use `.max` and `.min` on the chart window object, and they might expect it to work, but in this PR, it's either a no-op or an internal error.

PR #369, opened 2021-01-28
Stacked bar chart

JohnUUU:
# Notable Code Changes 

1.  We split the bar-chart-series into 2 different dataseries: 
  - **Bar-chart-series** (Bar charts and Frequency charts)
  - **Multi-bar-chart-series** (Grouped Bar charts and Stacked Bar charts). 

2. In order to support the axis-format function without having to rely on google bar chart's ISU format string [Which sometimes breaks in a way that we can't catch on Pyret's side]. The way we did it was we analyzed the axis patterns for different bar lengths [here](https://docs.google.com/spreadsheets/d/1vCeQlOGQv-wXw05vbpl4Zryu7XzkDn-BLWrDUjokGaI/edit?usp=sharing), and found out that google bar chart looks at the (bar length / 9) to determine the gridline step length. We used that to calculate and generate the axis. 

3. We added some more edge case error checks for grouped-bar-charts and bar-charts and now raise an error message  
- When the data is empty, the labels and values are of different lengths for bar-chart-series 
- When the data is empty, the data groups have different lengths, labels and values have different lengths,  and legends and values have different lengths. 

4. We implemented a way to create stacked-bar-charts namely 
``` 
from-list.stacked-bar-chart(
      labels :: List<String>, 
      value-lists :: List<List<Number>>, 
      legends :: List<String>) 
```
download (4)

# Methods Implemented
We implemented the following methods. 
### Tags:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[single] - Supported for bar-chart-series 
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[multi] - Supported for multi-bar-chart-series
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[issue] - Fails under some edge case or has some other issue

 
## Color Methods 
1. **Default Color** [single] - Changes the default color of all the bars in the series if another color is not specifically assigned to the bar. 
`bar-chart-series.default-color(c :: Color)`
Screen Shot 2021-01-27 at 9 58 03 PM

2. **Colors** [single][multi][issue] - Specifies the specific colors of each bar or each bar component 
`bar-chart-series.colors(loc :: List<Color>)`
Screen Shot 2021-01-27 at 10 02 24 PM

`multi-bar-chart-series.colors(loc :: List<Color>)`
Screen Shot 2021-01-27 at 10 04 58 PM
Screen Shot 2021-01-27 at 10 06 44 PM

## Sorting Methods 
1. **Sort-by** [single][multi] - Sorts the bars using a comparator and equality checker by bar height 
`bar-chart-series.sort-by(cmp :: (Number, Number -> Boolean), eq :: (Number, Number -> Boolean))`
Screen Shot 2021-01-27 at 10 15 34 PM

`multi-bar-chart-series.sort-by(cmp :: (Number, Number -> Boolean), eq :: (Number, Number -> Boolean))`
Screen Shot 2021-01-27 at 10 32 01 PM
Screen Shot 2021-01-27 at 10 30 00 PM

2. **Sort-by-label** [single][multi] - Sorts the bars using a comparator and equality checker by bar name
`bar-chart-series.sort-by-label(cmp :: (String, String -> Boolean), eq :: (Sring, String -> Boolean))`
Screen Shot 2021-01-27 at 10 22 12 PM

`multi-bar-chart-series.sort-by-label(cmp :: (String, String -> Boolean), eq :: (Sring, String -> Boolean))`
Screen Shot 2021-01-27 at 10 35 30 PM
Screen Shot 2021-01-27 at 10 36 12 PM

3. **Sort-by-data** [multi] - Scores the bars using a scorer that looks at all of the bar components. Then sorts the bars by score using a comparator and equality checker. Example is sorting by first bar component. 
```
multi-bar-chart-series.sort-by-data<A>(
      scorer :: (List<Number> -> A), 
      cmp :: (A, A -> Boolean), 
      eq :: (A, A -> Boolean))
```
Screen Shot 2021-01-27 at 10 40 56 PM
Screen Shot 2021-01-27 at 10 41 39 PM

## Axis Methods
1. **Add Pointers**[single][multi][issue] - Adds pointers at certain ticks on the axis to highlight certain lines of interest 
`bar-chart-series.add-pointers(ticks :: List<Number>, labels :: List<String>)`
Screen Shot 2021-01-27 at 10 52 04 PM

`multi-bar-chart-series.add-pointers(ticks :: List<Number>, labels :: List<String>)`
Screen Shot 2021-01-27 at 10 53 36 PM
Screen Shot 2021-01-27 at 10 54 30 PM

2. **Pointer Color**[single][multi] - Colors the pointers added by Add Pointers
`bar-chart-series.pointer-color(c :: Color)`
Screen Shot 2021-01-27 at 11 08 02 PM

`multi-bar-chart-series.pointer-color(c :: Color)`
Screen Shot 2021-01-27 at 11 10 25 PM
Screen Shot 2021-01-27 at 11 11 14 PM

3. **Format Axis**[single][multi] - Changes the names of the ticks on the axis 
`bar-chart-series.format-axis(f :: (Number -> String))`
Screen Shot 2021-01-27 at 11 16 39 PM

`multi-bar-chart-series.format-axis(f :: (Number -> String))`
Screen Shot 2021-01-27 at 11 17 48 PM
Screen Shot 2021-01-27 at 11 18 54 PM

## Scale Method 
1. **scale**[single][multi] - Scales the data by the function. Example is taking the num-log of the data. 
`bar-chart-series.scale(f :: (Number -> Number))`
Screen Shot 2021-01-27 at 11 22 07 PM

`multi-bar-chart-series.scale(f :: (Number -> Number))`
Screen Shot 2021-01-27 at 11 23 40 PM
Screen Shot 2021-01-27 at 11 25 03 PM

## Bar Type Methods
1. **Stacking-type**[multi] - Switches between stacking types. Stacking types are shown below. 
`multi-bar-chart-series.stacking-type(type :: String)`
Screen Shot 2021-01-28 at 12 40 35 AM

2. **Horizontal**[single][multi] - Switches between bar and column chart. Options are shown below. 
`bar-chart-series.horizontal(type :: Boolean)`
Screen Shot 2021-01-28 at 12 44 50 AM

`multi-bar-chart-series.horizontal(type :: Boolean)`
Screen Shot 2021-01-28 at 12 46 23 AM
Screen Shot 2021-01-28 at 12 48 19 AM

## Descriptor Object Methods 

1. **Annotations**[single][multi][issue] - Adds annotations that describe certain bars and bar-components 
`bar-chart-series.annotations(ann :: List<Option<String>>)`
Screen Shot 2021-01-27 at 11 35 08 PM

`multi-bar-chart-series.annotations(ann :: List<List<Option<String>>>)`
Screen Shot 2021-01-27 at 11 46 35 PM
Screen Shot 2021-01-27 at 11 49 29 PM

2. **Interval**[single][multi][issue] - Adds an interval at the top of a bar or bar component 
`bar-chart-series.intervals(int :: List<List<Number>>)`
Screen Shot 2021-01-28 at 12 05 01 AM

`multi-bar-chart-series.intervals(int :: List<List<List<Number>>>)`
Screen Shot 2021-01-28 at 12 09 53 AM
Screen Shot 2021-01-28 at 12 10 48 AM

3. **Error Bars**[single][multi][issue] - Adds an error-bar at the to of a bar or bar component
`bar-chart-series.error-bars(int :: List<List<Number>>)`
Screen Shot 2021-01-28 at 12 16 34 AM

`multi-bar-chart-series.error-bars(int :: List<List<List<Number>>>)`
Screen Shot 2021-01-28 at 12 17 00 AM
Screen Shot 2021-01-28 at 12 17 22 AM

4. **Interval-color**[single][multi] - Colors the intervals 
`bar-chart-series.interval-color(c :: Color)`
Screen Shot 2021-01-28 at 12 22 46 AM

`multi-bar-chart-series.interval-color(c :: Color)`
Screen Shot 2021-01-28 at 12 22 52 AM
Screen Shot 2021-01-28 at 12 21 26 AM

# Issues 
1. The current **colors method** doesn't support coloring bars in the middle or end bars without touching the beginning bars 
2. Multibars are not very compatible with annotations in some situations because it's prone to being clustered, overlapping, and stacked-bar-charts alternate drawing bar-components and annotations so a lot of the annotations get overlayed by a bar component and is unreadable. Same thing happens with intervals and error-bars for stacked-bar-charts
Screen Shot 2021-01-28 at 12 27 04 AM

3. Add-Pointers: Sometimes the pointers will be to close to one another and one of them will fail to be displayed
Screen Shot 2021-01-28 at 12 29 39 AM

4. Add-Pointers + Horizontal: The pointer names don't show up when the bar-chart is horizontal.
Screen Shot 2021-01-28 at 12 33 03 AM

5. Travis CI takes around 50 seconds to run the tests for chart.arr. The other chart test suites take around 5 seconds to run, so we might want to reduce the tests to the basics and not test edge cases.

# CI Error 
There's an inconsistent CI timeout error that we can't seem to track down but we know that our implementations have passed previous CI tests and that it consistently makes it past the chart-test file, At this point though, we might as well just push what we have first.


src/web/arr/trove/chart.arr
```diff
@@ -122,6 +285,367 @@ y-max-method = method(self, y-max :: Number):
...
+  scaled-tab = map(scale-row, list-of-rows) ^ builtins.raw-array-from-list
+  scaled-self = self.constr()(self.obj.{tab: scaled-tab})
+  scaled-values = map({(row): map(exact-sf, row ^ get-values)}, list-of-rows)
+  {max-positive-height; max-negative-height} = 
+    multi-prep-axis(scaled-self.obj!is-stacked, scaled-values)
+
+  scaled-self.make-axis(max-positive-height, max-negative-height)
+end
+
+stacking-type-method = method(self, stack-type :: String):
```
sorawee, line 449: Does it make more sense to define a new data definition, rather than to use a string, for `stack-type`?

src/web/arr/trove/color.arr
```diff
@@ -52,6 +52,7 @@ gold = color(255, 215, 0, 1)
 yellow = color(255, 255, 0, 1)
 olive = color(128, 128, 0, 1)
 burlywood = color(222, 184, 135, 1)
+amaranth = color(229, 43, 80, 1)
```
sorawee, line 55: Just to give you a headsup, this also needs an update at the `pyret-docs` repo, specifically on this file: https://github.com/brownplt/pyret-docs/blob/master/src/trove/color.js.rkt

src/web/js/trove/chart-lib.js
```diff
@@ -70,6 +71,28 @@
     return rgb2hex(IMAGE.colorString(checkColor(v)));
   }
 
+  function convertPointer(p) {
+    return {v: toFixnum(p.dict.value) , f: p.dict.label}
+  }
+
+  // Converts a list with func:
+  function convertListWith(func, l) {
```
sorawee, line 79: I think this is unsafe. JS Stack might overflow. I'm also not sure if using `val.dict.first` as a terminating condition is actually OK. In any case, I think there already exists a similar but safe function in core Pyret already, so you should just use it. Another possibility is to convert a list to a (raw-)array from Pyret side.

src/web/js/trove/chart-lib.js
```diff
@@ -269,22 +396,251 @@
...
+
+    // Adds each row of bar data and bar_color data
+    table.forEach(function (row, idx) {
+      const bar_color = idx < colors_list_length ? colors_list[idx] : default_color;
+      data.addRow([row[0], toFixnum(row[1]), bar_color]);
+    });
+    addAnnotations(data, rawData);
+    addIntervals(data, rawData);
+
+    var options = {
```
sorawee, line 474: Use `let` or `const` if possible. Also, there's a missing semicolon.

src/web/arr/trove/color.arr
blerner: Where did this color come from?  It's not an X11 color, or a DrRacket color, and I would like to have some sort of systematic reason for adding new colors...

src/web/arr/trove/chart.arr
JohnUUU: I think it makes some sense. 

The main difference between the two series is that 
- Bar-chart-series has its data represented by a List<Number> which contains the heights of the bars. 
- Multi-bar-chart-series has its data represented by a List<List<Number>> which contains the heights of the bars separated into components. 

The reason this is important is that many of the methods shown above must be implemented differently in each case. For example: 
- Colors is defined to color each bar in the bar-chart-series case and each component in the multi-bar-chart-series 
- Sort-by-data only makes sense in the context of a multi-bar-chart-series, where you might want to only sort by certain components 
- Calculating the axis depends on the highest bar displayed and the method needed to find the changes for each data-series. 
- Scale also has to be mapped differently based on each data-series 

Also stacking type would have to be able to change between the List<List<Number>> (multi-bar-chart-series) data and List<Number> (bar-chart-series) data. 
While we can switch from multi-bar-chart-series to bar-chart-series naturally,  It doesn’t seem feasible to go from bar-chart-series to multi-bar-chart-series naturally.

src/web/arr/trove/color.arr
JohnUUU: We mainly added this color because we wanted to test if we set up our gitHub branch and local pyret repl correctly. I guess after we confirmed that our commits actually result in changes to the local repl, we just moved on and left that in. 

The reason that we chose Amaranth is that we to quickly find a color that wasn’t already on the list of colors so we looked up the most underrated colors and picked one we liked from [here](https://www.mentalfloss.com/article/53121/11-colors-youve-probably-never-heard)

I’m perfectly okay with removing that color now if it’s unnecessary and goes against the current system of colors.

src/web/js/trove/chart-lib.js
JohnUUU: Got it! I’ll convert the lists into raw-arrays on Pyret's side in the next commit.

src/web/js/trove/chart-lib.js
JohnUUU: Got it! Will fix it in the next commit!

src/web/js/trove/chart-lib.js
```diff
@@ -70,6 +71,11 @@
     return rgb2hex(IMAGE.colorString(checkColor(v)));
   }
 
+  function convertPointer(p) {
+    return {v: toFixnum(p.dict.value) , f: p.dict.label}
```
sorawee, line 75: If I remember correctly, you shouldn't refer to `.dict` directly because it's an internal structure of Pyret objects that could potentially change in the future. Instead, use `get`.

JohnUUU: Sorry for the late reply, I was a little bit busy the last few days. I'll quickly swap out the .dict with get in the next commit!

src/web/arr/trove/chart.arr
```diff
@@ -75,13 +86,150 @@ fun to-table3(xs :: List<Any>, ys :: List<Any>, zs :: List<Any>) -> TableIntern:
...
+        pos-c = num-abs(c)
+        ask: 
+          | (pos-c > 0) and (pos-c < 1) then: recur(sci-notation(c * b, e - 1, b))
+          | (pos-c == 0) or ((pos-c >= 1) and (pos-c < b)) then: sci-notation(c, e, b)
+          | otherwise: recur(sci-notation(c / b, e + 1, b))
+        end
+    end
+  end
+  
+  {(n): recur(sci-notation(n, 0, base))}
```
blerner, line 157: Replace these lines with the more direct
```
  fun convert(n :: Number):
    if num-is-rational(n) and (n == 0): sci-notation(0, 0, base)
    else:
      pow = num-floor(num-log(num-abs(n)) / num-log(base))
      c = n / num-expt(base, pow)
      sci-notation(c, pow, base)
    end
  end
  convert
```

src/web/arr/trove/chart.arr
```diff
@@ -75,13 +86,150 @@ fun to-table3(xs :: List<Any>, ys :: List<Any>, zs :: List<Any>) -> TableIntern:
...
+      else: elm
+      end}, l.first, l)}
+
+  ask:
+    | is-stacked == 'none' then: 
+      # Find the tallest bar in the entire group 
+      positive-max-groups = map({(l): get-with-cmp({(a, b): a > b}, l)}, value-lists)
+      negative-max-groups = map({(l): get-with-cmp({(a, b): a < b}, l)}, value-lists)
+      max-positive-height = num-max(0, get-with-cmp({(a, b): a > b}, positive-max-groups))
+      max-negative-height = num-min(0, get-with-cmp({(a, b): a < b}, negative-max-groups))
```
blerner, line 207: replace these lambdas with `num-max` and `num-min`, and then eliminate `get-with-cmp` in favor of calling `fold` directly

src/web/arr/trove/chart.arr
```diff
@@ -75,13 +86,150 @@ fun to-table3(xs :: List<Any>, ys :: List<Any>, zs :: List<Any>) -> TableIntern:
...
+      negative-only-sum = {(l :: List<Number>): sum(filter({(e): e <= 0}, l))}
+      positive-sums = map(positive-only-sum, value-lists)
+      negative-sums = map(negative-only-sum, value-lists)
+      max-positive-height = num-max(0, get-with-cmp({(a, b): a > b}, positive-sums))
+      max-negative-height = num-min(0, get-with-cmp({(a, b): a < b}, negative-sums))
+      {max-positive-height; max-negative-height}
+
+    | otherwise: 
+      has-pos = any({(l): any( _ > 0, l)}, value-lists)
+      has-neg = any({(l): any( _ < 0, l)}, value-lists)
```
blerner, line 223: Be consistent in style: either ` > 0` or `{(e): e >= 0}`, but don't use both in such close proximity ;-)

src/web/arr/trove/chart.arr
```diff
@@ -122,6 +286,368 @@ y-max-method = method(self, y-max :: Number):
...
+  ticks = distinct(pos-ticks + neg-ticks) ^ builtins.raw-array-from-list
+  self.constr()(
+    self.obj.{axisdata: some(axis-data(axisTop, axisBottom, ticks))}
+    )
+end
+
+format-axis-data-method = method(self, format-func :: (Number -> String)):
+  cases (Option) self.obj!axisdata: 
+    | none => 
+      raise("Should never have reached this point. Yell at John for not setting up the axis properties somewhere where he should have and please report this as a bug")
```
blerner, line 417: Clean up this error message...

src/web/arr/trove/chart.arr
JohnUUU: Got it, replacing the function now!

src/web/arr/trove/chart.arr
JohnUUU: Good point, I guess this was an attempt at generalization that just ended up being unnecessary because we only needed num-min, num-max, and fold

src/web/arr/trove/chart.arr
JohnUUU: That’s my bad, making it more consistent now

src/web/arr/trove/chart.arr
JohnUUU: Yeah really sorry about this, I meant to go back and change it, Doing that now!

src/web/js/trove/chart-lib.js
```diff
@@ -174,6 +214,116 @@
 
   //////////////////////////////////////////////////////////////////////////////
 
+  /**
```
jpolitz, line 217: It's probably worth saying up top that this was copied from a jsfiddle more explicitly.

src/web/js/trove/chart-lib.js
```diff
@@ -174,6 +214,116 @@
...
+    colValues.forEach((row, rowN) => {
+        row.forEach((group, groupN) => {
+          group.forEach((val, i) => {
+          table.setValue(rowN, dataColNums[groupN] + i + 1, val);
+        });
+      })
+    });
+  }
+
+  /**
```
jpolitz, line 292: For example, I think we wrote this function, but the block above came from somewhere else – that's useful to know when debugging later.

src/web/js/trove/chart-lib.js
```diff
@@ -268,23 +418,259 @@
...
-        legend: {
-          position: isTrue(get(rawData, 'has-legend')) ? 'right' : 'none'
-        }
-      },
-      chartType: google.visualization.ColumnChart,
+      options: options,
+      chartType: horizontal ? google.visualization.BarChart : google.visualization.ColumnChart,
       onExit: defaultImageReturn,
-      mutators: [axesNameMutator, yAxisRangeMutator],
+      mutatorgraphs: [axesNameMutator, yAxisRangeMutator],
```
jpolitz, line 673: Why is this key changed? Don't see `mutatorgraphs` mentioned elsewhere in the file.

src/web/js/trove/chart-lib.js
```diff
@@ -781,7 +1167,7 @@ ${labelRow}`;
       }
 
       function setup(restarter) {
-        const tmp = f(globalOptions, rawData);
+        var tmp = f(globalOptions, rawData);
```
jpolitz, line 1170: Is this an important `const` -> `var` change?

src/web/js/trove/chart-lib.js
```diff
@@ -804,11 +1190,16 @@ ${labelRow}`;
...
+        // return true;
       }
 
       return RUNTIME.pauseStack(restarter => {
         google.charts.setOnLoadCallback(() => {
-          setup(restarter);
+          try{
+            setup(restarter)
+          } catch (e) {
+            return restarter.error(e);
```
jpolitz, line 1201: Thank you! It's so frustrating to debug when we miss errors like these.

src/web/js/trove/chart-lib.js
```diff
@@ -827,17 +1218,24 @@ ${labelRow}`;
...
+      'histogram': makeFunction(histogram),
+      'box-plot': makeFunction(boxPlot),
+      'plot': makeFunction(plot),
+    }, 
+    {
+      "LoC": ann("List<Color>", checkListWith(IMAGE.isColorOrColorString)),
+      "LoS": ann("List<String>", checkListWith(RUNTIME.isString)), 
+      "LoN": ann("List<Number>", checkListWith(RUNTIME.isNumber)),
+      "LoLoN": ann("List<List<Number>>", checkListWith(checkListWith(RUNTIME.isNumber))),
+      "LoLoLoN": ann("List<List<List<<Number>>>", checkListWith(checkListWith(checkListWith(RUNTIME.isNumber)))),
```
jpolitz, line 1235: Is the idea that the user will use these types?

jpolitz:
I pushed my review and merged so we can play with this on live horizon, but I would like answers to the questions I posed! :-)

src/web/js/trove/chart-lib.js
JohnUUU: The jsfiddle is actually from @eyanje so I'm not sure it's copied, probably more of an example. But it still might be worth putting up top for reference! Do you want me to do that in this PR or do you want to do it directly on horizon since this has already been merged. Sorry not 100% up to date on PR etiquette.

src/web/js/trove/chart-lib.js
JohnUUU: Yeah, but since the example is specific to the function, I think it actually makes sense to keep the jsfiddle comment above the function and maybe make a copy reference to it at the top if you want all the jsfiddles to be referenced at the top.

src/web/js/trove/chart-lib.js
JohnUUU: I'm going to be 100% honest. I have no idea why, I think this might have been a change made in one of the meetings we had with @blerner since I don't recall really ever changing this on my own, and I don't really remember what it does. I would check in with ben!

src/web/js/trove/chart-lib.js
JohnUUU: It doesn't seem to be, as far as I can tell the only things that we really edit are the variables of tmp which can still be changed if tmp is a const and tmp doesn't every really seem to be reassigned.

src/web/js/trove/chart-lib.js
JohnUUU: No problem! It's all thanks to ben lerner's help!

src/web/js/trove/chart-lib.js
JohnUUU: The main idea behind this was that pyret doesn't typecheck nested types so many of these methods that would require List<Type> or Option<Type> Etc... would give fairly unhelpful type errors when the wrong nested type was given. This is just so that we can give better type errors in that edge case!

src/web/arr/trove/chart.arr
sorawee: Hmm. I'm not sure if I understand your comment. What I meant is that we could define:

```
data StackType:
  | absolute
  | relative
  | percent
  | none
end
```

and use these instead of string.

JohnUUU: Oh I'm sorry I misunderstood what you saying. I think that forcing the users to create an new stacktype instance every time they want to change stacking types might be a bit too cumbersome and maybe too complicated/uncommon for anyone just starting out learning pyret. Thus personally I believe using strings would be better if only because they would be more familiar.

src/web/js/trove/chart-lib.js
sorawee: @blerner can you check this? It looks incorrect to me.

Here's a program that demonstrates the bug:

```

include chart

a-stacked-bar-chart-series = from-list.stacked-bar-chart(
  [list: 'CA', 'TX', 'NY', 'FL', 'IL', 'PA'],
  [list:
    [list: 2704659,4499890,2159981,3853788,10604510,8819342,4114496],
    [list: 2027307,3277946,1420518,2454721,7017731,5656528,2472223],
    [list: 1208495,2141490,1058031,1999120,5355235,5120254,2607672],
    [list: 1140516,1938695,925060,1607297,4782119,4746856,3187797],
    [list: 894368,1558919,725973,1311479,3596343,3239173,1575308],
    [list: 737462,1345341,679201,1203944,3157759,3414001,1910571]],
  [list:
    'Under 5 Years',
    '5 to 13 Years',
    '14 to 17 Years',
    '18 to 24 Years',
    '25 to 44 Years',
    '45 to 64 Years',
    '65 Years and Over'])


render-chart(a-stacked-bar-chart-series)
  .x-axis("hello")
  .display()
```

blerner: Joe and I saw that, and know it's wrong -- the question is whether it should be renamed back to `mutators`, or deleted entirely.  @JohnUUU needs to check it and see which behavior is intended.

src/web/arr/trove/chart.arr
sorawee: > I think that forcing the users to create an new stacktype instance every time they want to change stacking types might be a bit too cumbersome

I don't think it's any more cumbersome than creating stuff like `mode-solid` in the `image-typed` library.

> maybe too complicated/uncommon for anyone just starting out learning pyret

My goal is to provide a good interface for people to use. If some audiences need a very simple interface, wrapper libraries could be created for that purpose, and this is already the case for the chart library for Bootstrap.

`Option` looks pretty complicated compared to `null`. One can argue that people learning programming would understand `null` better, and that might even be true. Still, there is value in learning the right way to do things.

JohnUUU: Ah I see that makes sense, it's actually not cumbersome at all. Sorry, I must've gotten a bit confused from jumping around different coding languages, for some reason I thought you would have to make a new stacking type instance every time you called it like with classes. I actually have the opposite opinion now in that the strings might be unnecessary. Or should we do it the same was as Fillmode where we support both. If so should we split it into an typed and untyped library?

src/web/js/trove/chart-lib.js
JohnUUU: Yeah definitely, I'll do that right now!

src/web/arr/trove/chart.arr
sorawee: I think the fact that `Fillmode` is different in the typed and untyped library is mostly because the typed one is better and is what we want, but we can't remove the untyped one due to backward incompatibility. (the untyped interface has been around for a very long time)

One of the reasons why I am pushing to have `StackType` now is that if we wait longer and people start to use the current interface, then we won't be able to change the interface anymore due to the backward incompatibility constraint.

blerner: It's not quite that simple, either. In principle I agree with you, but in practice, using enumerated values doesn't necessarily make things simpler for teaching. Now you have a curricular dependency on explaining what these magic variable names mean, whereas the string-based API only requires teaching about strings.

I think possibly the right thing to do here is make *this* API be an enumeration, and ask @schanzer to wrap the API in a stringier version for use in bootstrap teachpack libraries. @schanzer and @jpolitz , agreed?

src/web/js/trove/chart-lib.js
JohnUUU: Oh I'm just realizing that new commits no longer get forwarded here since the PR has been merged. Should I open another PR? If you want to just make the change directly on horizon the right behavior happens when we rename it back to mutators.

blerner: Yes, go ahead and make a new PR and add me and Joe as reviewers.

PR #373, opened 2021-05-11
UI fonts and local setup details

asolove:
(Hi all! As I started exploring a few other projects, I noticed a couple small things. I'd normally send separate PRs but thought it would be wasted overhead in this case.)

1. Adds a CSS variable for the font stack for UI text, similar to the existing one for code. Then uses that in place of the hard-coded value in several places. Also applies that UI text font by default to the whole page, which catches some bits of text in tooltips and other places that were being shown with the browser-default serif font instead.
2. Adds a bit of detail to the README on how to get Google OAuth and file saving working locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,13 @@ At https://console.developers.google.com/project, make a project, then:
 
   Again, you should use `http://localhost:5000` as the referer for development.
 
+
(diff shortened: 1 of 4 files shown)


jpolitz (approved):
Looks great, all simple stuff and clear improvements.

PR #377, opened 2021-05-13
[WIP] Start restyling, working in the open

asolove:
⚠️ This PR is now a complete body of work to increase UI consistency and make CPO look more familiar to students who know Google Docs well. It's not every but of UI we might consider changing, but it's all the site-wide chrome bits that mostly didn't need any css/html changes. I still need to do a bit more browser and accessibility testing before being comfortable with it.

# Vision
[See discussion and mockups](https://docs.google.com/document/d/1diXEgx7TnF_KHnioZAxl458xx2IlBJ_kxG4G0s4mIjs/edit#bookmark=id.ptyfrlsvsbl5) from an earlier discussion.

# Milestones
I intend to propose incremental improvements in self-contained and easy-to-review PRs.

This first PR will propose some relatively safe, global, CSS-only changes:
  - [x] Title bar and menu styling
  - [x] Modal styling and visual hierarchy of cancel/submit buttons
  - [x] Run/stop button styling
  - [x] Clarify shadows and boundaries between menu, editor, and repl
  - [x] Clean up duplicate and inconsistent font rules

Future PRs will try to make more local changes.

# Preview
[Demo video](https://www.youtube.com/watch?v=MCHlEL6yLv8)

### DIFF
--- a/src/web/css/editor.css
+++ b/src/web/css/editor.css
@@ -945,7 +945,7 @@ body.smallHeader .replContainer {
 }
 
 #bonniemenuContents {
-  min-width: 200px;
+  min-width: 250px;
 }
 #filemenuContents {
   min-width: 200px;
@@ -956,7 +956,9 @@ body.smallHeader .replContainer {
   color: #333;
   border: 1px solid white;
   border-radius: 3px;
-  margin-top: -2px;
+  border-top-left-radius: 0px;
+  margin-top: -3px;
+  margin-left: 1px;
   background-color: var(--menu-contents-bg);
   box-shadow: 0px 1px 5px var(--shadow-color);
   overflow:hidden;
@@ -982,7 +984,7 @@ body.smallHeader .replContainer {
 }
 
 .menuButton > *, .menuContents > * {
-  padding: 8px 24px 6px;
+  padding: 10px 30px 10px;
   display: block;
 }
 
@@ -2096,10 +2098,13 @@ div.choiceContainer {
 div.choiceContainer input[type=text] {
   padding: 0.5em;
   font-size: 100%;
+  border: 1px solid #999;
+  border-radius: 2px;
 }
 div.choiceContainer .textLabel {
   margin-right: 1em;
 }
+
 div.choiceContainer > button {
   margin: 1em;
   cursor: pointer;
--- a/src/web/css/editor.css
+++ b/src/web/css/editor.css
@@ -23,7 +23,6 @@
   - box-shadow on .dropdown:hover (aside run button)
   - bg on .modal-header
   - bg on .modal-footer 
-  - bg on .menuContents div:not(.disabled):not(.inert):hover
   - box-shadow on #runButton, #breakButton (also on hover)
   - box-shadow on #run-dropdown-content li:hover
   - box-shadow on table.pyret-table thead
@@ -96,7 +95,7 @@
   --notif-area-bg: #111111;
 
   /* - bg on New/Save/Download/etc menu (.menuContents) */
-  --menu-contents-bg: #EFEFEF;
+  --menu-contents-bg: #fff;
 
   /*
   - border around expanded objects (.replOutput.expanded > dl) */
@@ -108,7 +107,7 @@
 
   /* 
   - text in .menuButton and .menuContents */
-  --menu-text: #000;
+  --menu-text: #333;
 
   /*
   - bg on .info-btn (question mark) in Bonnie menu */
@@ -954,16 +953,20 @@ body.smallHeader .replContainer {
 #filemenuContents .hidden { display: none; }
 
 .menuContents {
+  color: #333;
+  border: 1px solid white;
+  border-radius: 3px;
+  margin-top: -2px;
   background-color: var(--menu-contents-bg);
-  border: none;
-  box-shadow: 0px 2px 3px var(--shadow-color);
+  box-shadow: 0px 1px 5px var(--shadow-color);
   overflow:hidden;
   max-width: min-content;
+  padding: 6px 0;
 }
 
 .menuContents div:not(.disabled):not(.inert):hover{
   transition: box-shadow 0.2s;
-  background-color: var(--shadow-color);
+  background-color: #e4e4e4;
 }
 
 .menuContents div.disabled * {
@@ -979,7 +982,7 @@ body.smallHeader .replContainer {
 }
 
 .menuButton > *, .menuContents > * {
-  padding: 10px;
+  padding: 8px 24px 6px;
   display: block;
 }
 
@@ -2088,7 +2091,7 @@ table.choiceContainer {
    margin: 1em;
 }
 div.choiceContainer {
-   text-align: center;
+   text-align: left;
 }
 div.choiceContainer input[type=text] {
   padding: 0.5em;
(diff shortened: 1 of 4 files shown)


schanzer:
This is really exciting! Tracking for accessibility reasons.￼

asolove:
Hey all! I think the current state of this PR is all the scope that is worth doing at one time. I've done a first round of browser testing and just need to poke at accessibility some more. (Do you have any preexisting documents on how you think about accessibility for CPO? I see a lot of attention paid to ARIA, which I haven't touched, but some of the color contrast could maybe use a bit of a nudge.) 

At this point I'd encourage folks to give feedback after watching [a short demo video](https://www.youtube.com/watch?v=MCHlEL6yLv8) or running it themselves. 

Obviously I have never had to use this editor in a classroom setting and have probably missed things, from very small to fundamental principles. So please send nitpicks or complete teardowns! You won't hurt my feelings.

schanzer:
This video is damned impressive - kudos!

Re: a11y -- I put together a set of notes a few years ago. They're still largely accurate, and short enough that they're worth a quick read. Happy to talk in more detail if you like.

I take it the combo run/stop button is out of scope for this PR, and I like the more noticeable visual indicators you've added for now! Just want to make sure it isn't being lost. :) 

Speaking of visual indicators, here are a few things that I've always thought would be great to have:

- Definitions and Interactions areas display their names simply by being moused-over (see an example on www.WeScheme.org/openEditor)
- Definitions area or bottom status bar indicates whether it is in a dirty and/or un-save-able state. In other words, a teacher should know just by glancing at a kid's screen whether closing the browser window is going to result in all of their work disappearing forever. Having the status bar turn slightly red, for example, and appending "changes will not be saved" to the "Programming as XYZ" string would be a huge usability win.
- The same logic (`[CodeMirror.isClean](https://codemirror.net/doc/manual.html#isClean)`?) that tells us if there are unsaved changes could easily be used to highlight the "Run" button, as a subtle hint that it needs to be clicked. Just an idea.

asolove:
How sure we folks that the notion of "Save" is appropriate here? My sense is kids are reasonably used to apps just auto-saving rather than them having to manage that state. To avoid cases where something gets messed up, we could plausibly persist the undo history across sessions so they can also move backwards to previous states of the file.

schanzer:
If a student is not logged in, or they are viewing a published/read-only file, _there is no auto-save available_. Right now we provide no visual cue.

asolove:
Ah, yes! I would love to have auto-save where it is possible and then a very visible marker for unsaved or in a stat that's unable to save, maybe with a more visible-promo to make your own copy. Maybe make the definition window uneditable until you make a copy?

asolove:
@schanzer: and yeah, have not forgotten the combo run/stop button, but it's going to require rearranging a bunch of JavaScript that currently has hard-coded assumptions about the buttons. But I've got the styling working!

image

shriram:
I'm probably the old-fashioned fuddy-duddy here (I am!), but having a Save
button is reassuring because sometimes I really want to make sure something
was saved, irrespective of the "automatic" behavior.

Whether or not students have such an expectation, many of our users are
teachers, and non-CS teachers at that, who are more used to tools like
Word—so taking away a Save would, I think, weird them out.

jpolitz:
Beautiful stuff! To be expedient, I'm going to say that this LGTM and merge it. It's way easier for e.g. @schanzer and @shriram to try things once they show up on horizon and give better feedback. I don't see any changes that are hard to back out, and I like how things look in your video demo :-)

(I'm afraid a longstanding TODO of mine is to set up Heroku to spin up instances for PRs)

asolove:
Ok, let me know if folks have feedback from Horizon!

asolove:
@jpolitz: looks like [the merge build failed](https://travis-ci.com/github/brownplt/code.pyret.org/builds/226260218). Is that normal? Something worth looking into?

jpolitz:
Yeah I _think_ I fixed it now. It was a “real”-ish issue unrelated to this. We have integration tests that load things from our curriculum, and one of the curriculum files changed so the test didn't work anymore.

(At first I worried it might be that a Selenium script was brittle-y relying on some classname you changed, but that's not the case).

shriram:
Is this on Horizon yet? I'm not sure https://pyret-horizon.herokuapp.com/editor looks different…

asolove:
Nope. Looks like Horizon hasn't been deployed in 10 days:
https://github.com/brownplt/code.pyret.org/deployments/activity_log?environment=pyret-horizon

jpolitz:
Yeah there’s a flaky test blocking the deploy. I’ve been trying to fix it
off and on when I have spare time this week, still haven’t had success.

asolove:
I've been meaning to get a Sauce account working, so I'll see if I can get that running and poke at the test of you don't get to it first.

jpolitz:
FWIW, this shouldn't need to involve Sauce, as what's failing is failing on
Chrome headless that's running directly in Travis.

The tests that are having issues fail intermittently and (for these) I
can't reproduce it on my own, I just see it in Travis. They all load a
#share URL, click the run button, and enter something at the REPL.

There have been a few issues with tests like these over the years, often
indicating an actual bug, but sometimes I can't figure out if there's a
race in waiting for the REPL to show up and testing for its contents, or
network failures in Travis, or some other kind of intermittent issue that
doesn't reflect a real bug.

asolove:
For folks following this, [Horizon](https://pyret-horizon.herokuapp.com) is now updated with the styling changes.

schanzer:
Totally awesome. There are some a11y issues (which also exist in master), but that's probably better handled in a separate issue/PR.

asolove:
Ooh can you share what you have in mind? I know about some issues with the
modals and the drop down menus, but I’d love to know what you have in mind
so I can look at that, too.

schanzer:
We can discuss in more detail later, but right off the bat the following important use-case is broken:

- User hits F6 to trigger the focus carousel, and keeps hitting it until the toolbar at the top has focus (mostly works now, except the toolbar never gets focus. Instead, only the first menu (Bonnie) does.
- User hits Tab to navigate to the first menu (Bonnie), which causes menu label to receive focus and the menu to become visible (doesn't right now)
- User hits Down-arrow to navigate the menu (doesn't work right now)
- User hits ESC to close the menu and return to the menu label

sorawee:
This looks really awesome!

Following are problems I noticed (well, these were problems in master too).

The dropdown menus are weird regarding deactivation. For example, in the "Run" menu (which has "Run" and "Type-check and run"), clicking the "down triangle" again doesn't deactivate the menu. Clicking empty space in the top bar doesn't deactivate the menu. You need to either click the editor or a clickable button to make it deactivate.

Perhaps in the main menu, "Log out" should be greyed out if the user is programming as a guest.

The font size controls are greyed out, making it look like the controls are disabled. However, they actually are not.

The line height in the REPL is too tall in every theme except the default theme (Ensign).

The "Contribute detailed usage information" icon is too big.

asolove:
Thank you both for sharing!

The menu issues (including the run dropdown menu) are on my list to tackle
next, along with changing some menu organization and consistently showing
keyboard shortcuts next to menu items, where supported.

I hadn't actually tried toggling themes to see if there were major styling
differences, so that's a great call-out. We probably want to remove as much
layout as possible from the themes and have them just be controlling colors.

schanzer:
Additional nit: 

In the editor, a guest user sees a button called "Connect to Google Drive".
On the homepage, a guest user sees a button called "Log In".

Both of these buttons have the exact same functionality, but two completely different titles. I'd really like them both to say "Log In".

shriram:
Nits aside, OMG, this is looking so beautiful!

Just wondering, do we want the "chrome" to change when we pick a different color scheme?

sorawee:
> The line height in the REPL is too tall in every theme except the default theme (Ensign).

Actually this one might be something that @thomascastleman can help?

Screen Shot 2021-06-28 at 5 24 53 PM

vs

Screen Shot 2021-06-28 at 5 25 07 PM

thomascastleman:
Oddly, I am unable to reproduce this on Brave/Firefox on Linux (I just checked on a Mac though and am seeing this. There's also weirdness when toggling between themes). 

Regardless, if it's happening in all except the default, the culprit is likely something in [`non-default-overrides.css`](https://github.com/brownplt/code.pyret.org/blob/horizon/src/web/css/themes/non-default-overrides.css), which applies styling to themes that aren't the default (and is also the only place where not just colors are modified). 

It's not clear to me what's doing this. @asolove is definitely better versed in this kind of problem than me, so it'd be great to see if he has any ideas about this.

Issue #350, opened 2020-10-08
better labels


shriram:
The Rename menu option looks like

![image](https://user-images.githubusercontent.com/75030/95524890-99854f00-09a0-11eb-842c-2faed52c08d4.png)

Both buttons look wrong, at least if I understand the semantics correctly: "Close" should be "Cancel" and "Submit" (which has a needless BSDM ring to it) should be "Rename" (one could image more precisely alternatives like "Proceed" or "Commit" or something, but I imagine that for kids, "Rename" is the easiest to understand).

shriram:
(Close should be Cancel because Close doesn't make clear whether the renaming happened anyway or not.)

schanzer:
While we're at it...
![image](https://user-images.githubusercontent.com/77126/95637120-f34c4e80-0a5e-11eb-9666-37f9f962f198.png)

"Close" should be renamed "Cancel"

Once you share a file for the first time, the "Update" button is a little confusing ("what am I updating...?"). And of course, it's extra confusing that clicking _either button_ has the same effect! Why even make two buttons?

asolove:
Would folks be interested in a PR to update the modal buttons more broadly? Right now they're always static for every modal, so I'd start by letting `cpo/modal-prompt/Prompt` accept new options for the button text.

I might also suggest restyling the left button to not be red, as it's generally the safer/do-nothing option, rather than something dangerous enough it should be styled red. (That said, I know users and documentation may have come to rely on details of this UI, so if touching it is dangerous, wanted to check first.)

asolove:
I don't want to be presumptuous, but if I'm thinking about this anyways, is there any interest in a broader UI refresh? I took [some notes on bits of the UI that seemed unfamiliar](https://docs.google.com/document/d/1diXEgx7TnF_KHnioZAxl458xx2IlBJ_kxG4G0s4mIjs/edit#bookmark=id.ptyfrlsvsbl5) and had an idea to update the menus, dialogs and top buttons at the same time. But this was just relative to my own understanding and I know a lot more research would be needed to understand the issues that real users face and try to solve them as part of that effort.

jpolitz:
That document and your thoughts are not presumptuous, they are spectacular! Thank you so much for writing all of this down, what a gift! I put some comments in the doc.

I think that:

[...]

PR #379, opened 2021-05-28
Improve UI for save as, rename, and import dialogs.

asolove:
This PR improves UI styling, a11y, visual hierarchy, and some micro text for three of the CPO modal dialogs.

(After starting work on other modals, they require enough thinking that they need to be in separate PRs.)

# Scope

### Global dialog improvements
- [x] Dialogs can customize the text on the close button, default to "Cancel" (#350)
- [x] Focus the first focusable element after shown.
- [x] Allow dialogs to opt into narrower style.

### Individual dialogs
- [x] Save As / Rename
  - [x] Change submit button text to "Save"
  - [x] Narrower, clearer spacing
  - [ ] Suggest better/funnier name for unnamed files? Change "Untitled" suggestion to "avast.arr"?

- [x] Import image
  - [x] Clarify styling of the three separate options in the table
  - [x] Make code samples smarter, adjusting to number of files selected
  - [x] Find a better name for import image style picker. "Import type" could be mistaken for a PL "type". Need to convey "how do you want these images to be imported?"
  - [ ] Make more of each row clickable to select the radio button

# Preview 

## Save As / Rename

Production:
image

With this PR:
image

## Import image

**Production:**
image
(Code samples show like this, even when only 1 or 2 files selected.)

**With this PR:**
Code samples change depending on number of files selected.
- e.g. with one file:
image
- with three files:
image

### DIFF
--- a/src/web/css/editor.css
+++ b/src/web/css/editor.css
@@ -2100,13 +2100,21 @@ table.choiceContainer {
 }
 div.choiceContainer {
    text-align: left;
+   margin-top: 16px;
 }
 div.choiceContainer input[type=text] {
+  display: block;
+  margin-top: 10px;
+  width: 100%;
+  box-sizing: border-box;
+  padding: 10px 12px;
+  font-family: var(--font-stack);
   padding: 0.5em;
   font-size: 100%;
-  border: 1px solid #999;
-  border-radius: 2px;
+  border: 1px solid #aaa;
+  border-radius: 3px;
 }
+
 div.choiceContainer .textLabel {
   margin-right: 1em;
 }
@@ -2194,7 +2202,7 @@ table.pyret-row {
 }
 
 .modal-content.narrow {
-  max-width: 500px;
+  max-width: 400px;
 }
 
 /* Add Animation */
@@ -2301,7 +2309,7 @@ table.pyret-row {
 }
 
 .modal-footer {
-  padding: 2px 20px 20px;
+  padding: 20px;
   display: flex;
   flex-direction: row-reverse;
 }
--- a/src/web/css/editor.css
+++ b/src/web/css/editor.css
@@ -2096,8 +2096,42 @@ table {
 
 table.choiceContainer {
    border: 0px;
-   margin: 1em;
+   margin: 0px;
+   background: none;
+   border-radius: 4px;
 }
+
+table.choiceContainer tr {
+  background: none !important;
+}
+table.choiceContainer td {
+  border-bottom: 1px solid #ccc;
+}
+table.choiceContainer tr:last-child td {
+  border-bottom-width: 0px;
+}
+table.choiceContainer .pyret-modal-option-example {
+  padding-right: 0px;
+}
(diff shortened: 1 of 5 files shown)


jpolitz:
I like all the changes above!

I don't have data on this, but I would bet a small amount that the number people who have used the `[list: ...]` option in that import dialog is in the single digits. The first two are by far the most common. I suspect we could eliminate that option to simplify things.

sorawee:
The list with one element option also looks kinda useless.

shriram:
Wow. Who knew this could be this nice.

asolove:
I was going to ask about usage of the import options and any feedback folks had heard.

I agree the single file case looks a bit odd right now, but here's the weird chain of thinking that led me there:

- From one standpoint, we could just skip this whole step and always import as values. Students might benefit from choosing a meaningful variable name, as leaving around `img1` seems like an avoidable antipattern. This would certainly make things faster/simpler.
- But given the editor's lack of multi-cursor or regex support (which is understandable for this audience!), if you really _did_ want to import a list of 50 images, it would be painful to import them on separate lines and have to go back to add commas and indentation manually. So the list option is useful.
- The first several time I used the modal, I didn't really internalize that I could import multiple images at the same time. I only noticed this on the nth try, after stopping to think about why anyone would ever choose the "list" option. If we removed the list option when one file was selected, you might _never_ discover that you could select multiple, and that would be sad. (There may also be a strong argument for completeness here: to help students navigate the classic set-theory dilemma about distinguishing an item from the set/list which contains just that item.)

All of this thinking in circles led me to offering all options all the time, though some of them are more sensible than others. Moving further in this direction, we might also consider inserting single image values at the current cursor location rather than a new line, if you want to specifically import the file inside an existing expression.

But I could see an equally-strong argument for just removing this step and always using values and letting users figure out variable names or additional structure themselves.

asolove:
Note: after starting to think about the tricky bits of the publish dialog, I'm going to cut scope of this one to just the current set (save as, rename, and insert image) and start the publish work in another issue.

shriram:
Just a comment that this is a good argument for leaving the list import option. I also think it has the benefit of inspiring ideas — e.g., you *realize* you can list-import, so you create a bunch of images to import that way (for instance, to make a zoetrope). TBH, I had forgotten/not realized you can do a list import until I saw the GUI proposal…

jpolitz:
Fair enough @shriram . I guess list is handy if you navigate to a directory, cmd/shift select a bunch. Doing values and inserting the [list: ..., ..., ...] would be annoying because of all the commas. Probably really magical for those <10 people :-)

asolove:
Hey all! The perfect is the enemy of the good, and I think this branch is now a clear enough improvement over the current state that I'd like to get it into Horizon.

Let me know if y'all have any objections or suggestions.

jpolitz:
I'm going to merge and so we can poke at it on horizon and give any feedback after. This is a pretty clear improvement, thanks!

PR #383, opened 2021-06-30
prevent scrollbars from causing REPL line height issues

thomascastleman:
This addresses the line height problems mentioned in #377 by changing scrollbars in the non-default themes. 

Previously, they used the `::webkit-scrollbar` selectors to create a custom scrollbar for both the REPL and CodeMirror windows. We've switched to using the [`color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme) property, which is only affecting the color of the scrollbar and seems reasonably well-supported.

We have checked Firefox, Brave, Chrome and Safari and so far the worst effect is that in Safari when switching between themes, the scrollbar maintains its previous color until you click on it. I would argue this is a better issue than line height problems that prevent the REPL from being readable.

### DIFF
--- a/src/web/css/themes/base16.css
+++ b/src/web/css/themes/base16.css
@@ -165,9 +165,7 @@ body.base16 {
   --loader-text:        var(--white);             /* text containing loading messages during page load */
 
   /* Scrollbars */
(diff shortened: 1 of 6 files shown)


jpolitz:
Agreed that this is an improvement, thanks!

PR #386, opened 2021-07-13
Fix bug with editor/repl height

### LINKED ISSUE #384: resize with Bonnie menu open causes editor to start halfway down screen
To reproduce:

1. Open the Bonnie menu
2. Click and drag the middle bar between definitions/interactions windows to resize

The top of the editor and REPL are now positioned just below where the Bonnie menu ends. Resizing again with the menu closed causes them to return to normal.

asolove:
This addresses the issue raised in #384.

The minimal change here was to use `offsetHeight` instead of `scrollHeight` so we no longer count the height of the absolutely-positioned menu.

A major question I have is in what case this code is needed at all. Seems like the same goal could likely be met in CSS and I don't understand in what case the toolbar is larger than the CSS size and requires running this code dynamically on all scroll/resize events. If someone can explain that, I'd be happy to take a crack at a more-complete solution.

### DIFF
--- a/src/web/js/beforePyret.js
+++ b/src/web/js/beforePyret.js
@@ -723,7 +723,7 @@ $(function() {
   }
 
   function updateEditorHeight() {
-    var toolbarHeight = document.getElementById('topTierUl').scrollHeight;
+    var toolbarHeight = document.getElementById('topTierUl').offsetHeight;
     // gets bumped to 67 on initial resize perturbation, but actual value is indeed 40
     if (toolbarHeight < 80) toolbarHeight = 40;
     toolbarHeight += 'px';


schanzer:
Oh dear god yes!!

jpolitz:
@ds26gte do you have any more context/memory of the issue or how to reproduce why we needed this?

https://github.com/brownplt/code.pyret.org/commit/a365632a56f1659d1e7bb88624f08c67647fb2e8

blerner:
I vaguely recall that CodeMirror tends to not to like to extend to the full vertical height of its container, and occasionally Shenanigans must be employed to convince it to stay the height of the window -- I don't think a pure-CSS solution sufficed, and the CM state would become stale.  *If* I'm recalling that correctly, then this code makes sense and then flipping scrollHeight to offsetHeight makes sense as well...

jpolitz:
Merging because it's obviously a fix, so no reason to wait.

Issue #382, opened 2021-06-28
Design a theme that is friendlier to color-blindness


blerner:
Now that we have theming support, it should be easier to manage this.  The distinctions between blues/purples, or purple/red, seem not to have been visible enough to some teachers.  I'll try to get some more feedback from teachers which ones they had the most trouble discerning.

Pinging @thomascastleman (for themes), @schanzer and @ds26gte (for accessibility)

schanzer:
Worth looping in @asolove as well, since I know this is something he's thinking about in his UI revamp.

thomascastleman:
Great--with the theming we have it should be relatively straightforward to create a new theme (just make a copy of [the template](https://github.com/brownplt/code.pyret.org/blob/horizon/src/web/css/themes/template.css) and choose new colors). 

We could also start from another theme already in use and work off of that, which might make more sense if only certain parts are in need of color changes.

blerner:
A suggestion was made by one teacher to rely more heavily on font changes, rather than solely color changes. For instance, our incomplete and complete string literals are distinguishable even with red/green color blindness, because they also change boldness. We should consider bold, italic, and possibly font-size changes as well.

Issue #326, opened 2020-03-25
Make definitions and interactions have a white background


schanzer:
Taking screenshots of REPL output for use in lesson plans is _painful_, because there's always a light gray background that needs to be scrubbed out. This is insanely inconvenient for anyone preparing lesson materials, slides, etc.

blerner:
I *really* don't want to do this -- the bright nearly-white is glaring enough, and I'd rather it were dimmer, honestly!  What I'd prefer, though, is if we supported two or three stylesheets (maybe available in the Bonnie menu), for "default", "dark mode", and "high-contrast / teacher mode", where the latter is best for projectors (with frequently crappy bulbs and low contrast) or for this use-case of screenshots.

shriram:
Doesn't the light grey background actually help, by making clear this is a
specialized context? Essentially like "quotation marks"?

>

schanzer:
No. The gray doesn't help. There's an implicit assumption among young people (and some old people!) that the absence of color is _white_. And while that is totally nuts, it's also both understandable and a constraint that Bootstrap has to live with (or devote lesson time to correcting).

I'll be honest - this tripped me up when I was writing a lesson yesterday. I had an outline circle that looked almost white in CPO, and I copy/pasted it into an image file with a white background. For a brief moment, I looked back at CPO to see if I'd accidentally made the image gray! And if it messed me up, it's likely to trip up teachers and students.

But my chief complaint isn't really that it's potentially confusing. It's just that it makes copy/pasting REPL output into usable slides kind of a nightmare. I like @blerner 's suggestion of multiple style sheets, but given that there are multiple issues left over from CSPdWeek _last year_ I'm concerned about making what would be a one-line CSS fix into a years-long engineering task to support multiple stylesheets.

jpolitz:
Can you give an example of a problematic screenshot? Is it that you want to pair expressions with output, so right clicking and saving the image doesn't work? I usually right-click and save as, which should have the right transparency for any background.

blerner:
I'm almost inclined to make this be a _new theme_, in line with the work on #329 #332 , which would allow us to customize things even further for "Screen sharing" mode: we could have punchier contrast in colors to avoid washed out projector displays, larger fonts, whatever ideas we want...

schanzer:
I'm happy with the theme work serving as a solution to this issue. Closing.

PR #390, opened 2021-08-09
Color blindness support

thomascastleman:
This begins to address #382 by extending themes to support control over font styling and other non-color aspects of the editor that can be used to create contrast where only color is being used currently. It may also address #326.

There are a few changes:
- I eliminated the distinction between the default theme (Ensign) and all the other themes--default was relying on certain styling in editor.css that was being overriden by all other themes. Now, all the themes truly are controlling the same styles and there's no need for special overrides.
- I added theme variables for controlling non-color font-styling on Pyret tokens (see [`src/web/css/themes/README.md`](https://github.com/thomascastleman/code.pyret.org/tree/color-blindness-support/src/web/css/themes) for more info), so we can now do things like make all types bold, etc.
- I added a new theme, High Contrast Light, which is a prototype of using these new features to make elements of the syntax highlighting/editor more visible to color-blindness/low contrast.

Here are some screenshots of the new high contrast theme to show what kinds of things can be styled with these changes:

high-contrast-code

high-contrast-checks

## Comparison 
The check block colors may seem like a striking difference but they were chosen to increase color contrast for various colorblindness types. See the below comparison (colorblindness simulated with the [Spectrum Chrome extension](https://chrome.google.com/webstore/detail/spectrum/ofclemegkcmilinpcimpjkfhjfgmhieb?hl=en)).

|                   | Ensign  | High Contrast Light |
|  ------------- | ------------- | ------------- |
| Protanopia |  ensign-protanopia | high-contrast-protanopia |
| Deuteranopia | ensign-deuteranopia | high-contrast-deuteranopia |
| Tritanopia | ensign-tritanopia  | high-contrast-tritanopia |

Ultimately this is a work in progress and these are just the first ideas I had. I wanted to get feedback given the new changes to themes that allow us to style more aspects of the editor. I think it can be hard to tell what kind of font changes would be helpful vs. distracting.


jpolitz:
Wow, this is terrific! I don't see a reason not to merge this sooner rather than later.

I could imagine finding a less Watermelon Sugar High palette that still works across the different constraints, but I don't know how to find it off the top of my head, and this is a real need.

shriram:
Ditto. This is amazing!

thomascastleman:
Good to hear! I've made a few updates:

- I changed the green color for check success into a light blue color which does an even better job of providing color contrast between the failing/erroring colors. If anyone is strongly opposed to using a blue for success, I could change it back but it really does a good job of contrasting:

|   | High Contrast Light (with blue) |
| ------------- | ------------- |
| Normal  | high-contrast-with-blue |
| Protanopia | protanopia  |
| Deuteranopia | deuteranopia |
| Tritanopia | tritanopia |

- I also added a High Contrast Dark theme, which applies many of the same contrasting colors and non-color font styling as High Contrast Light, but is a dark theme. Here are some samples:

  high-contrast-dark-definitions
  high-contrast-dark-interactions

  It uses the same blue color to provide good contrast in check block results.

Let me know what you think!

blerner:
I _think_ the white-on-pink text for "some tests in this block may not have run" might fall afoul of minimum contrast standards (because the background is too bright) -- worth double checking.  Actually, in most of those screenshots, the white text is hard to read, and black would be better.

Possibly-useful tip: Firefox's developer tools has an Accessibility tab that will let you simulate these color-deficiencies as well, and also will check for contrast issues.  Not sure if the Spectrum extension can do the latter, or if Chrome has contrast-checking tools for you to use...

thomascastleman:
Oh nice, I didn't know Firefox had that built-in--thanks! 

Using the Firefox checker for contrast issues, it didn't seem to flag any violations for the white-on-pink, but I agree it seems to be cutting it close. I changed both themes to use black text there and here's what it looks like under low-contrast simulation:

light-low-contrast

dark-low-contrast

thomascastleman:
@blerner @jpolitz I've made Ben's updates--is there anything else you all would like to see changed? If not, do you think it's ready to release with the summer update? Thanks!

blerner:
One more bit of feedback, from one of our color-vision-impaired teachers:
> I can easily distinguish almost all the colors in both themes. I like them a lot. The conditionals in the light theme (lines 8 and 9) are basically black to me. In the dark theme they look great. In the dark theme, the two floats at the end of line 31 look the same to me, but they look different in the light theme. Otherwise, I find all the colors much much more helpful. And I love the feedback table colors.

I clarified further: the exact-nums on lines 8,9 are a bit dark and read mostly as black, as compared to the roughnums on lines 30,31.  And the booleans on lines 8,9 are too dark a red and are seen mostly as black.  The request is to brighten up the red of the booleans, and to tweak the colors of numbers: exactnums should be blue `hsl(240, 100%, 39%)` (which is the current roughnum color), and roughnums should be `hsl(200, 80%, 40%)` (which is between roughnums' color and strings' color `hsl(164, 100%, 31%)`) -- I've confirmed with the teacher that those proposed colors are distinguishable from each other and from strings.

jpolitz:
Re: summer update – I think this will actually be nice to release separately. It's also low-risk (whereas the summer update was risky; lots of stuff affected).

It's also a nice, separate thing to announce on our road to better accessibility. So I think we can put this out at any time once we're satisfied (and we seem to be a commit or two away from that).

thomascastleman:
Sounds good, that makes sense.

@blerner Here are booleans adjusted to be brighter:

booleans

and the number color changes applied to the light theme:

nums-light

Are the number changes intended for the dark theme as well? Applying them yields quite dark exact nums which I find a bit hard to read against the background:

nums-dark

I can get something more readable by boosting the lightness on each to about 70%:

dark-nums-higher-lightness

and it still seems to maintain pretty good contrast.

thomascastleman:
If it would help, we could also add some non-color styling to rough-nums, like underline/italics/bold if that would help the distinction be unmissable.

blerner:
No, the teacher's comments were specifically about the light theme; the colors you originally had in the dark theme were pretty good.  I might separate the hues of the two original dark-theme number colors you'd had, making one of them greener like we just did with the light theme.

thomascastleman:
Here's the dark theme with a more green roughnum color:

nums-dark-separated

jpolitz:
LGTM!

blerner:
Can we lift this up to main and get it deployed?

jpolitz:
Wanted to note that this is on production now!

image

PR #393, opened 2021-08-12
Performance: combine scripts used in CPO editor

asolove:
This PR builds many js files loaded by the CPO editor into a single, much smaller, file.

# Why?
CPO needs to be faster! This PR is one small step towards that.
- One way of looking at it is that this PR replaces 21 network calls for 660kb with 1 call for 360kb, a savings of 95% and 45%, respectively.
- Another, more accurate, way of looking at it is that this PR reduces the total network calls to boot Pyret from ~100 to ~80 and payload size from 33.3mb to just 33mb, which is only a savings of 20% and 1%.

We have a lot more work to do.

Nevertheless, this change is not just window dressing! Because the browser has a limit on parallel network requests, these 21 requests were delaying it from starting to request the main pyret jarr file. So even though they are relatively much smaller, they were blocking the beginning of the main factor slowing down CPO. Combining them is therefore a relatively efficient way to get the large jarr file loading much earlier (perhaps 4s earlier on high-latency connections).

Because we currently re-request cached files to do cache validation checking with the server, this change also speeds up the experience by even more in the case where you already have all the code cached locally. (We should change the way that works, so we version the js urls and can check cache locally, but that is another day and another PR)

# How
JavaScript is a dangerous place to do this kind of global restructuring. I was too afraid of making accidental semantic changes to read through all the code impacted and try to do something smart, like turning it all into modules and compiling it with webpack.

Instead, I am trying to do the least-risky and least-content-aware transformation I can. So this PR:
- takes the existing js files, using their existing build outputs in the `Makefile`
- concatenates them together in the exact order they were loaded by the html (I have left out several other files which were intermingled with code loaded from other domains, because I didn't want to worry about the semantics of reordering or take on the job of bundling the external files)
- use uglifyjs to do a minimally-semantics-aware minification of the result (e.g. no dead code elimination or renaming of top-level variables)

There was one case where I _did_ have to touch an individual JS file to make this strategy work. The `log.js` file was previously not valid JS itself, but instead used templating to insert a few environment variables (`LOG_URL`, `GIT_REV`, `GIT_BRANCH`) that it depended it. This prevented it from being parsed by uglify and required it to be processed specially by the node server on each request. I replaced the use of templating with global variables that are now set in each html file that uses them.

# Review notes
This change could potentially break _literally anything_ about CPO. While I am testing things I know about and haven't yet run into problems, I'd appreciate suggestions of fragile or little-used corners of the JS code that we should try out to check that things are safe.

To review the PR, I recommend looking at the changes in their conceptual order, which is not actually the order of the git commits:
1. Pass the additional environment variables to the templating of the [editor](https://github.com/brownplt/code.pyret.org/pull/393/files#diff-79f4c7cdc0b026f6e556196f5911a449157807cffdb3f9d33f0b5404b4ffc2f3L330-R324) and [index](https://github.com/brownplt/code.pyret.org/pull/393/files#diff-79f4c7cdc0b026f6e556196f5911a449157807cffdb3f9d33f0b5404b4ffc2f3R108-R111) html files, and then define them as JS global variables in the [editor](https://github.com/brownplt/code.pyret.org/pull/393/files#diff-2f66053f6892a14a729e6f47dfab983c2be8382cb1b0f3e39f1362ec8778a362R418-R425) and [index](https://github.com/brownplt/code.pyret.org/pull/393/files#diff-2929665d1098d9b7f4e87dfe5ea5ae21baa7554de1e1739f574d36d6225a5858R18-R24) html files.
2. Rewrite the [portions of `log.js` that used templating](https://github.com/brownplt/code.pyret.org/pull/393/files#diff-36c69b23fe7810c1aac1a13c63a5daf01237a2fcb7dce878d184a52e6828e5e4L98-R103) to instead use the global variables. Remove the [special processing](https://github.com/brownplt/code.pyret.org/pull/393/files#diff-79f4c7cdc0b026f6e556196f5911a449157807cffdb3f9d33f0b5404b4ffc2f3L92-L103) for `log.js` in the server.
3. Define the [list of files to be combined](https://github.com/brownplt/code.pyret.org/pull/393/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R190-R213), add a [new makefile target](https://github.com/brownplt/code.pyret.org/pull/393/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R212-R213) to process them, and then call that from the [current build targets](https://github.com/brownplt/code.pyret.org/pull/393/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L236-L239).
4. Replace [lots of script tags with a single one](https://github.com/brownplt/code.pyret.org/pull/393/files#diff-2f66053f6892a14a729e6f47dfab983c2be8382cb1b0f3e39f1362ec8778a362L421-L443) for the new combined file.
5. (Side quest) Fix some [whitespace and consistency issues in the Makefile](https://github.com/brownplt/code.pyret.org/pull/393/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L169-R188) since we're there anyways
6. (Side quest) Remove the unused [`IndexedDBBackend`](https://github.com/brownplt/code.pyret.org/pull/393/files#diff-36c69b23fe7810c1aac1a13c63a5daf01237a2fcb7dce878d184a52e6828e5e4L21-L34) from `log.js` and then [stop loading](https://github.com/brownplt/code.pyret.org/pull/393/files#diff-2f66053f6892a14a729e6f47dfab983c2be8382cb1b0f3e39f1362ec8778a362L421) its `dexie` dependency.
7. (Side quest) Remove a [commented out script tag](https://github.com/brownplt/code.pyret.org/pull/393/files#diff-2f66053f6892a14a729e6f47dfab983c2be8382cb1b0f3e39f1362ec8778a362L469-L470) for loading Caja. RIP Caja, we know that you have gone on to a better place, where you can play with Google Reader and look down and smile on SES.

# Next steps
Once this is tested and landed, next steps on performance are split between small and large work.

Small work, small benefit:
- Kick off network request for `cpo-main.jarr` earlier in the life of the page. (link rel=prefetch?)
- Minify other bits of js by running uglifyjs over them (especially `beforePyret.js`), which should save another 400kb in transfer size
- Add gzip encoding option for all js files, saving another 300kb

Scary work, potentially huge benefit:
- Audit the contents of the cpo-main.jarr bundle. I hope to find a few categories of things that can be removed:
  - npm dependencies that are bundled with it but not actually needed in the browser environment
  - metadata and sourcemap information for compiler internals that aren't needed in the browser environment


jpolitz:
Hey @jswrenn – if you have a minute could you scan the changes to log.js and see if there's any special reason they might break? I know there's a sort of tortuous path on production through Heroku config settings to those variables and want to make sure there's nothing special in your memory that could break here.

jpolitz:
All the Selenium tests pass. I'm going to go ahead and merge this so we can see what it looks like live on horizon. Other than logging JS, the only other thing to test is the auth* JavaScript, which is easier to test there anyway.

PR #395, opened 2021-08-16
Performance: remove unused optional dependencies of node modules.

asolove:
Several of the node packages we bundle have _optional_ dependencies that are currently being bundled, but not used, taking up ~700kb of space in the built version of `npm-dependencies.js` that gets packaged with `cpo-main.jarr`.

This change directs browserify to explicitly not bundle those packages.

Examples:
- `colorspaces` is a 10kb library for color math. It [_optionally_ pulls in stylus](https://github.com/boronine/colorspaces.js/blob/master/colorspaces.js#L324), in order to offer some helpers to people who use it. We currently bundle the whole stylus compiler code (400kb) and never use it.
- `js-md5` and `js-sha256` are small, pure-js implementations of their respective functions. But [they](https://github.com/emn178/js-md5/blob/0263ab835f4847351a710f98488203108c0e1e90/src/md5.js#L155) [also](https://github.com/emn178/js-sha256/blob/189bb9b03782b80e59516dfbea78f16b5d9754ce/src/sha256.js#L83) optionally pull in node modules that offer faster versions of their features when run on the server. As a result, we are currently bundling node's `crypto` and `buffer` implementations and not using them. (~300kb).

Because these optional imports are in the source of dependencies, I was unable to remove them surgically and instead have just instructed browserify to _never_ include them. If we ever _do_ want to use those packages, we'll need to undo this change. (But, also, we have probably made a mistake.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -253,7 +253,8 @@ libpyret:
 	$(MAKE) phaseA -C pyret/
 
 $(BUNDLED_DEPS): src/scripts/npm-dependencies.js
-	node_modules/.bin/browserify src/scripts/npm-dependencies.js -o $(BUNDLED_DEPS)
+	# Explicitly exclude crypto, buffer, and stylus, nested npm dependencies that aren't needed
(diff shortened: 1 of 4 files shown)


jpolitz:
This looks great!

PR #396, opened 2021-08-16
Performance: strip JavaScript comments from cpo-main

asolove:
⚠️ Because this branch needs `uglifyjs`, it is built on top of the work #393. That PR should be reviewed and land before this one. The only changes to review here are in 40497f04f391c317580aa14408d2610c32e8c1c1.

I'm still figuring out the whole build process, but somehow the `cpo-main.jarr` file contains the raw text of some JavaScript packages we depend on, like `source-map`. By including the raw source, we are even including somewhere around 3mb of inline documentation from these packages. By running uglify over the source, we remove these unused bits.


asolove:
NB: just rebased this on latest horizon to simplify the diff and poke CI to run again.

jpolitz:
Is this worth merging on its own or should I hold off until https://github.com/brownplt/code.pyret.org/pull/397 is more settled?

asolove:
This is worth something like a 5% savings by itself.

PR #406, opened 2021-10-01
fix table borders in Firefox

thomascastleman:
For whatever reason tables have a rectangular border in Firefox.

with-border

I only noticed because I accidentally set the table border to be black for the default theme. This PR removes that border so that tables are consistent across browsers, but still allows themes to control the internal borders between table cells.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "code.pyret.org",
       "version": "0.0.0",
       "hasInstallScript": true,
       "license": "Apache-2.0",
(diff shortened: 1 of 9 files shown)


blerner:
Are the `package-lock.json` changes necessary for this PR?  If not, remove them...

I'm not sure whether we want this change exactly this way, or not -- my initial aesthetic reaction was "hmm, I think I do want a border, but I want it to match the rounded corners of the table..."  @jpolitz or others, thoughts?

blerner:
Closing this PR, since there's a cleaner (1-line!) CSS fix to make the borders hug the table more tightly.  This will also give tight borders to rendered rows.

PR #411, opened 2021-12-02
Fix #408: hide legend if there is only one DataSeries

### LINKED ISSUE #408: Hide "plot 1" if a ChartWindow contains only one DataSeries
Bootstrap:Data Science has _tons_ of scatter plots, but all of them have only a single DataSeries. That makes the image below slightly confusing for kids and teachers alike - especially if they print out a bunch of plots, and they all say "plot 1"!

image

Can this be removed, if there's just one?

sorawee:


blerner:
LGTM

PR #415, opened 2022-01-06
[VMT] hacky focus fix

anzook:
A very hacky fix for the interaction panel stealing focus on the receipt of an interaction event from another user in VMT.

Currently setting the CodeMirror repl-promt class as readOnly nocursor (no focusing)

### DIFF
--- a/src/web/js/events.js
+++ b/src/web/js/events.js
@@ -3,27 +3,37 @@ let targetOrigin = POSTMESSAGE_ORIGIN;
 function commSetup(config, messageCallback) {
   function sendEvent(data) {
     console.log("Sending from CPO ", event);
-    config.sendPort.postMessage({
-      protocol: 'pyret',
-      messageNumber: counter++,
-      timestamp: window.performance.now(),
-      data: data
-    }, targetOrigin);
+    config.sendPort.postMessage(
+      {
+        protocol: "pyret",
+        messageNumber: counter++,
+        timestamp: window.performance.now(),
+        data: data,
+      },
+      targetOrigin
+    );
   }
-  config.receivePort.onmessage = function(event) {
-    if(typeof event.data === 'string' && event.data.indexOf("setImmediate") === 0) { return; }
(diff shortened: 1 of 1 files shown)


jpolitz:
Some more commentary – I'm pretty sure this has to be handled Pyret-side because focus is so hard to manage in the browser, and CPO does a programmatic focus of the next interaction after each interaction runs.

This hack has some nice properties – it doesn't require plumbing through any flags about synthetic events, etc. It clearly happens around the extent of the run of interactions and "undoes" itself.

However, it's super brittle DOM-wise. So maybe the best thing to do is add an extra entrypoint on `runner` that toggles whether the next interaction should be `nocursor` or not.

Also if @blerner or someone knows a way to make it so a surrounding page can stop elements in an iframe from getting focused, this could be solved (and usefully solved for other apps) VMT-side.

src/web/js/events.js
```diff
@@ -86,11 +96,20 @@ function makeEvents(config) {
         const interactions = message.currentState.interactionsSinceLastRun;
         const src = interactions[interactions.length - 1];
         interactionsSinceLastRun.push(src);
-        const dontFocus = (e) => { e.preventDefault(); e.stopPropagation(); };
-        $(window).on("focusin", dontFocus);
+        const dontFocus = (e) => {
+          e.preventDefault();
+          e.stopPropagation();
+        };
```
jpolitz: `dontFocus` isn't used anymore, can remove.

blerner:
Pretty certain a surrounding page can't affect the dispatch process of events within an iframe, since that would break most of the security properties of iframes...

src/web/js/events.js
anzook: Good catch, thanks Joe.

I removed this and the commented code

anzook:
Updating the code led to unnecessary commits- if this is merged, I recommend either 'squash and merge' or I can rebase to a single commit.

jpolitz:
Merging to test live.

PR #421, opened 2022-01-28
Smaller sbc continuation

JohnUUU:
This is a smaller Pull Request with just the colors method and sort methods implemented for pie charts. 
Also includes small adjustments talked about earlier like having an actual enumerated StackType and having colors be connected to the data rather than the index.


src/web/arr/trove/chart.arr
```diff
@@ -86,6 +92,10 @@ fun to-table3(xs :: List<Any>, ys :: List<Any>, zs :: List<Any>) -> TableIntern:
   map3({(x, y, z): [raw-array: x, y, z]}, xs, ys, zs) ^ builtins.raw-array-from-list
 end
 
+fun to-table4(xs :: List<Any>, ys :: List<Any>, zs :: List<Any>, ks :: List<Any>) -> TableIntern:
+  map4({(x, y, z, k): [raw-array: x, y, z, k]}, xs, ys, zs, ks) ^ builtins.raw-array-from-list
```
blerner: Instead of the lambda here, you can just write `raw-array.make4`, which is what that particular 4-argument call to `[raw-array: ...]` does.

src/web/arr/trove/chart.arr
```diff
@@ -163,6 +173,17 @@ where:
 |#
 end
 
+fun string-to-stacktype(s :: String) -> StackType: 
+  doc: ```Converts ['none', 'absolute', relative', percent'] to their respective stacktype```
+  ask: 
+    | string-equal(s, 'none') then: grouped
```
blerner, line 187: This feels weird, in a "One of these things is not like the others!" sort of way.  Why shouldn't this string be "grouped"?

src/web/arr/trove/chart.arr
```diff
@@ -172,14 +193,14 @@ fun prep-axis(values :: P.LoN) -> {Number; Number}:
   {max-positive-height; max-negative-height}
 end
 
-fun multi-prep-axis(is-stacked :: String, value-lists :: P.LoLoN) 
+fun multi-prep-axis(is-stacked :: StackType, value-lists :: P.LoLoN)
```
blerner: `is-stacked` sounds like a boolean flag.  This parameter should be named `stack-type` or something.

src/web/arr/trove/chart.arr
```diff
@@ -1215,7 +1243,7 @@ fun exploding-pie-chart-from-list(
   offsets.each(check-num)
   labels.each(check-string)
   default-pie-chart-series.{
-    tab: to-table3(labels, values, offsets)
+    tab: to-table4(labels, values, offsets, range(0, labels.length()))
```
blerner: Since `to-table3` is your custom function, and you're always passing in `range(0, labels.length())`, try using [`map3_n`](https://github.com/brownplt/pyret-lang/blob/862baff55029a75b99806833dedd11291805feaf/src/arr/trove/lists.arr#L560) instead, which will produce those indices for you.

src/web/arr/trove/chart.arr
```diff
@@ -1264,7 +1292,7 @@ fun bar-chart-from-list(labels :: P.LoS, values :: P.LoN) -> DataSeries block:
   {max-positive-height; max-negative-height} = prep-axis(rational-values)
 
   data-series = default-bar-chart-series.{
-    tab: to-table2(labels, rational-values),
+    tab: to-table3(labels, rational-values, range(0, labels.length())),
```
blerner: Ditto, use `map2_n`

src/web/js/trove/chart-lib.js
```diff
@@ -401,14 +401,29 @@
 
   function pieChart(globalOptions, rawData) {
     const table = get(rawData, 'tab');
+    const default_colors = ['#3366CC', '#DC3912', '#FF9900', '#109618', '#990099',
```
blerner: Leave a comment in the code saying you copied these from Google's default colors, and leave a link to there, so we can find them again.

blerner (changes requested):
Some quick, fairly-small readability changes

src/web/arr/trove/chart.arr
JohnUUU: Yeah sorry about that the reason why it's "none" is because that's the string we give to Google for it to parse whether or not it's stacked or grouped ("none" vs "stacked"). I changed it to grouped on pyret's side since I thought that would make a bit more sense but the string we give google charts is still "none".

src/web/arr/trove/chart.arr
JohnUUU: Yeah sorry about that, the parameter for google charts was is-stacked so I just copied that over. Changed it to stack-type!

src/web/arr/trove/chart.arr
blerner: Ok -- does this string ever appear to users of this code?  Or is it entirely internal to your implementation?  If it's internal, I don't worry about it as much.

JohnUUU: The string is entirely internal, externally the users input one of the enumerated stacktype: 
[grouped, absolute, relative, percent]

jpolitz:
We think this code looks pretty good – do you have a sense of what we should do to test it/mark it as complete @JohnUUU ?

JohnUUU:
Thanks, most of the edge case tests I could think up of I've put in bar-chart-test.arr and I've done most of the pairwise combination tests for these graphs already. But those 2 approaches are the only real main approaches to testing I've done for these so if there's anything else that's Important to test for before marking it as complete I would do that! Otherwise if it seems well tested enough and works for your purposes easily enough then I think it would be fine to mark it as complete!

PR #434, opened 2022-05-01
Fix font size menu item

asolove:
This PR does a few small tweaks to improve the font size menu item:

2022-04-30 21 46 47 localhost 2e56d3a8d36f

- Changes the menu item to styled more like normal items, as previous gray color led some users to believe it was disabled/inaccessible, and rearrange buttons closer to each other so experimentation is faster.
- Uses flexbox layout rather than some fixed widths based on occasionally-wrong math to avoid a few cases where the menu wrapped onto multiple lines.
- Fixes a bug where Safari sometimes showed long decimal font sizes (e.g. "10.1094854857px")
- Changes +/- from divs to buttons, so they are slightly more accessible for folks using assistive technology. (Though not massively, and I expect to try to fix some larger menu a11y issues soon)

Unrelated to this core change, and possibly cherry-pickable into its own branch, I have turned off the cacheing of css and html files that was being done by the mustache template engine, which meant you sometimes had to restart the whole server to see html changes, even after rebuilding assets manually.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -82,6 +82,7 @@ function start(config, onServerReady) {
   app.engine('html', mustache());
   app.engine('js', mustache());
   app.set('view engine', ['html', 'js']);
+  app.set('view cache', process.env.NODE_ENV !== 'development');
 
   app.get("/current-version", function(req, res) {
     res.status(200);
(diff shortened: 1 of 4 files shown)


jpolitz:
> Unrelated to this core change, and possibly cherry-pickable into its own branch, I have turned off the cacheing of css and html files that was being done by the mustache template engine, which meant you sometimes had to restart the whole server to see html changes, even after rebuilding assets manually.

Ah, nice. That's smart and something I never bothered to look up. I don't mind that being part of this, thanks for explaining.

This looks generally great to me, going to merge.

PR #435, opened 2022-05-11
horizon -> mainmast

jpolitz:
Small horizon updates (font size, heroku config stuff, APP_ID for picker)


jpolitz:
NB: there might be another update to `.travis.yml` coming for new release-specific Heroku tokens. Some of the tweaking above is to avoid putting horizon-specific configuration onto the release branch

PR #452, opened 2022-08-29
Adding commenting and searching functionality to CPO

jchen:
# Introduction
This adds editor functionality based on brownplt/pyret-codemirror-mode#10 (and depends on this PR). Added are the following keybinds:
* <kbd>Control/Cmd</kbd>-<kbd>/</kbd> to comment lines or multiple lines. 
* <kbd>Control/Cmd</kbd>-<kbd>F</kbd> and <kbd>Option</kbd>-<kbd>Control/Cmd</kbd>-<kbd>/</kbd> to find and replace respectively. 

# Testing
Tested by running locally. Also feel free to bump line 53 of `package.json`:
```
 - "pyret-codemirror-mode": "git://github.com/brownplt/pyret-codemirror-mode.git#master",
 + "pyret-codemirror-mode": "git://github.com/jchen/pyret-codemirror-mode.git#master",
```
to use Codemirror mode changes from brownplt/pyret-codemirror-mode#10 atop CPO to test new functionality.

### DIFF
--- a/src/web/js/beforePyret.js
+++ b/src/web/js/beforePyret.js
@@ -183,6 +183,9 @@ $(function() {
       rulersMinCol = CODE_LINE_WIDTH;
     }
 
+    const mac = CodeMirror.keyMap.default === CodeMirror.keyMap.macDefault;
+    const modifier = mac ? "Cmd" : "Ctrl";
+
     var cmOptions = {
       extraKeys: CodeMirror.normalizeKeyMap({
         "Shift-Enter": function(cm) { runFun(cm.getValue()); },
@@ -194,7 +197,8 @@ $(function() {
         "Esc Right": "goForwardSexp",
         "Alt-Right": "goForwardSexp",
         "Ctrl-Left": "goBackwardToken",
-        "Ctrl-Right": "goForwardToken"
+        "Ctrl-Right": "goForwardToken",
+        [`${modifier}-/`]: "toggleComment",
       }),
       indentUnit: 2,
       tabSize: 2,
--- a/Makefile
+++ b/Makefile
@@ -156,6 +156,9 @@ build/web/js/foldcode.js: $(CM)/addon/fold/foldcode.js
 build/web/js/foldgutter.js: $(CM)/addon/fold/foldgutter.js
 	cp $< $@
 
+build/web/js/comment.js: $(CM)/addon/comment/comment.js
+	cp $< $@
+
 build/web/js/pyret-mode.js: $(PYRET_MODE)/mode/pyret.js
 	cp $< $@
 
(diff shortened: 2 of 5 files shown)


asolove:
Oh, this is cool! Please add to the shortcuts dialog as well so they are discoverable.

src/web/js/beforePyret.js
```diff
@@ -183,6 +183,9 @@ $(function() {
       rulersMinCol = CODE_LINE_WIDTH;
     }
 
+    const mac = CodeMirror.keyMap.default === CodeMirror.keyMap.macDefault;
```
jpolitz, line 186: Oh neat, didn't know this existed.

src/web/js/beforePyret.js
```diff
@@ -194,7 +197,8 @@ $(function() {
         "Esc Right": "goForwardSexp",
         "Alt-Right": "goForwardSexp",
         "Ctrl-Left": "goBackwardToken",
-        "Ctrl-Right": "goForwardToken"
+        "Ctrl-Right": "goForwardToken",
+        [`${modifier}-/`]: "toggleComment",
```
jpolitz, line 201: Ah OK no change for `{modifier}` here, which is least surprise. Thanks!

jpolitz (commented):
All seems solid to me! I like @asolove 's comment, that's the only change I'd make. Thanks for following our very-manual makefile strategy for these files 😬

src/web/js/beforePyret.js
jchen: Yup - wanted to retain the same functionality for all other commands (on a mac, Cmd-Right/Left) are still start/end of line.

jchen:
Added all the shortcut dialogs! I didn't know this help menu was a thing so I did learn something new! 

While I was in there, I styled the help dialog a bit - hope that's reasonable: 
image

jpolitz:
OK couple comments now that I've played with this:

I'd like to disable it in REPL entries. These can get focused for lots of reasons and it looks super weird there and completely occludes code in single-line entries 
image

Is there a way to get a graphical/clickable close button for it? Many users will not think to press escape *while focused on the search box* especially if they brought it up by accident by fat-fingering Ctrl-V

jchen:
@jpolitz These both should be possible - I might need to tinker with it a bit. 

I suppose an unintended consequence of importing the script globally was that it applies to all CodeMirror instances which I didn't realize. 

Feel free to revert the PR if this blocks other horizon changes from getting merged.

jpolitz:
Nah I don't think it's worth reverting! It's a good improvement for a lot of people. We should just improve it.

PR #456, opened 2022-09-21
add image- enhanced versions of histograms, bar- and pie-charts

schanzer:
This largely follows the behavior of image-scatter-plot, so there shouldn't be (m)any surprises. A few things that I'm not sure about, however:

- I'm making the assumption that an image display for histograms, bar and pie charts will only ever be constructed from a single series. This is the same assumption made in image-scatter-plot as well, but I don't know enough to be confident that it's safe to make it here

Thanks in advance for taking a look!

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
   "author": "Joe Politz",
   "license": "Apache-2.0",
   "devDependencies": {
-    "chromedriver": "^97.0.0",
+    "chromedriver": "^106.0.0",
     "selenium-webdriver": "^3.6.0",
     "webpack-cli": "^3.1.2"
   }
--- a/src/google-auth.js
+++ b/src/google-auth.js
@@ -12,7 +12,7 @@ var FULL_OAUTH_SCOPES = [
                     "email",
                     "https://www.googleapis.com/auth/spreadsheets",
                     "https://www.googleapis.com/auth/drive.file",
-                    "https://www.googleapis.com/auth/drive.readonly",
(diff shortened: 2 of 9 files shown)


schanzer:
@jpolitz Now that I've figured out the lingering issues, I've made sure all the tests pass and updated to the latest from horizon. Hopefully this is ready to merge.

When running the full build, it looks like Travis hangs when processing pie-charts. This doesn't happen locally, though, and I don't know how to address it

jpolitz:
Did you fix the issue with the rogue `Either`?

I'm seeing this error hang the tests but want to make sure I'm on the right version:

image

schanzer:
Huh. I _thought_ I had, since I was able to run the test suite locally. Is it possible that a test would hang on one machine but not on another?

jpolitz:
Absolutely it's possible. I'm getting identical results on a Macbook and my mac Mini. This is both with Chrome (via chromedriver), and copying the `test-util/pyret-programs/charts/image-scatter-and-line-test.arr` file and running in Firefox.

image

jpolitz:
To be clear, `image-bar-chart-test` is failing (that may just be a test that needs to be updated; it's comparing a transparent pixel to a red one, see screenshot below).

But `image-scatter-and-line-test.arr` is hanging with the Either error (well, waiting for a very long timeout).

jpolitz:
I'm going to merge this because the tests _do_ appear to be working, and then work out ChromeDriver on horizon.

PR #474, opened 2023-07-13
Dropdowns for context selection

westluke:
For the dashboard page of CPO, turns the "open editor" button into a split button with dropdown that lets you choose a context to open the new file with.

Also changes the "choose context" modal in the editor to give users the option to choose from whitelisted contexts.

The whitelisted contexts, as well as the default context, get stored in environment variables in `.env`

This also bumps the version of the react dependency so we can use `React.Fragment`

The styling on the dropdowns and the modal is really really minimal, I was told to let someone else handle that


jpolitz:
Merging so we can test it out on horizon; I haven't reviewed in a ton of detail yet, but seeing it live will be useful.

jpolitz:
Here's what I'm seeing.

On `/editor`:

image

image

On `/`:

image

image


It seems like `process.env` is empty in these contexts. I did a fresh clone and copied `.env.example` to `.env`. We _will_ need to add these new env options to Heroku to deploy, but that's not part of the Selenium build.

I wonder if there's something different about your build (webpack or similar) that's templating into different places differently?

jpolitz:
Suggestion: Explicitly template into `index` and into `editor` in `server.js` rather than relying on the magic of `process.env` with webpack/react builds, e.g.

https://github.com/brownplt/code.pyret.org/blob/horizon/src/server.js#L108

(Will need to also add a variable declaration in the appropriate HTML template file)

PR #477, opened 2023-08-22
Pcardune blocks

jpolitz:
This is the continuation of the *amazing* work from https://github.com/brownplt/code.pyret.org/pull/462.

I finally had time to dig into this with @schanzer over the weekend – this is awesome!

I think this is worth putting on horizon at the very least.

Since the original PR:

- Fixed various (re)sizing and z-index issues
- Got Gdrive saving working reasonably
- Added blocks for functions


jpolitz:
Cool stuff:

image

shriram:
This is indeed awesome!

I just wish the function call block weren't called "apply"? That has too
much Scheme baggage. Why not just "call"?

jpolitz:
Could delete the keyword altogether and have it just look like `name(arg)`, too.

Kinda depends what math teachers benefit the most from seeing, I think.

jpolitz:
For those curious, this is what it's like to add blocks:

https://github.com/brownplt/code.pyret.org/pull/477/commits/b863fbf7d88d94c77d25c2dd086e54e81571bf9f

Many, many tricks to try and learn for layout, naming, etc here, but this is pretty tractable to muck with. I found editing the XML directly to be just fine (and Copilot is _very_ good at helping with typing here)

shriram:
> Could delete the keyword altogether and have it just look like name(arg), too.

Oh, sure, that too. @schanzer?

schanzer:
Oooooh, I never considered copilot for this. Nice tip, @jpolitz !

@shriram 110% agree that `name(arg)` is the right way to go here.

PR #480, opened 2023-09-02
Fix #472: repeating decimal overline bar rendering

### LINKED ISSUE #472: Overline bars on repeating decimals don't show on small-size chromebooks
From Eric Allatta:

`On chromebooks some teachers could not see the repeating decimal bar until they increased the pyret font zoom above 14. `

We may need to _get_ a small chromebook to test this, but my hunch is that subpixel rendering is turned off. The solution might be to [take more control over the line thickness](https://css-tricks.com/css-underlines-are-too-thin-and-too-low-in-chrome/)....

asolove:
In #472, there was speculation the overline bar was rendering too thin. But the real issue is that the `.trace` container is cutting off vertical overflow, and current math means only a small fraction of the overline is visible, which gets not rendered on some display sizes/configurations.

This is a very conservative change, to just include an extra pixel of top padding to let the overline render. (There are more principled ways to do this, but without knowing what bad cases the `.trace {overflow-x: auto; }` was solving, I didn't want to touch them.)

### DIFF
--- a/src/web/css/editor.css
+++ b/src/web/css/editor.css
@@ -477,6 +477,7 @@ div.CodeMirror span.CodeMirror-nonmatchingbracket-region {
 
 div.trace {
   overflow-x: auto;
+  padding-top: 1px;
 }
 div.trace.error {
    background-color: var(--trace-err-bg);


schanzer:
@jpolitz @blerner any chance of getting this merged and up on CPO? The mysteriously-disappearing overline is a huge PITA for a lot of teachers

Issue #442, opened 2022-07-12
Can "Run" and "Stop" shrink when the screen is narrow?


schanzer:
Every time I run PD over Zoom, I am reminded that the Stop button goes rogue and covers part of the Definitions Area when the screen is narrow. 

Can we use a CSS media query to check for super-narrow screens, and shrink the width of those buttons accordingly?

asolove:
Yeah, I will find a way to do something about this. The main issue here is the Run/Type dropdown reserving enough space for the very wide text for the typed label.

PR #481, opened 2023-09-02
smol mode fix

asolove:
Fix for #442.

image

UI previously broke at widths <700px. Now works down to ~550px. The run and break buttons are rendered narrower under 700px.

We should do a more principled cleanup to make the menu layout work nicely on mobile devices, but this is easy/fast.


blerner:
This looks easy enough to me.

Issue #486, opened 2023-09-30
Chart package: residuals


schanzer:
### Note: Algebra 2 is blocked on this feature

Long story short: Bootstrap's algebra 2 materials rely on the ability to quantitatively compare the fitness of different models, in order to motivate students to use quadratics, exponentials, logarithms, etc to find the best-possible fit. We originally relied on R^2 for this, since it's the only quantitative comparator we have. @shriram  and Nancy Pfenning correctly pointed out that in doing so, _we've essentially set kids up to become bad data scientists_: "find the best R^2 you have and you've got a correlation!"

We need a different quantitative measure of fitness that doesn't immediately leap to value judgements about correlations, and Nancy proposed _residuals_ as the way to do it. I think this is spot-on, as it both solves all of our immediate problems for Algebra 2 and opens up important content we'd like to cover anyway in Data Science.

jpolitz:
Notes from a recent meeting – we need the ability to have something that gives custom labels on the scatter dots (right now they'd just show the actual number, not the residual, if you used https://pyret.org/docs/latest/chart.html#%28part._chart_render-charts%29).

So some help from the JavaScript chart library would be needed here.

jpolitz:
The implementation of `residuals` is straightforward:

```
fun residuals(xs, ys, f):
  for map2(x from xs, y from ys):
    f(x) - y
  end
end
```

jpolitz:
@ds26gte this PR is a useful place where we added a new chart type/primitive:

https://github.com/brownplt/code.pyret.org/pull/456

jpolitz:
This single commit may be useful to see: https://github.com/brownplt/code.pyret.org/pull/456/commits/e637c3dee1b1056bbd5245323128439608c1248c

Also this PR for box-plot: https://github.com/brownplt/code.pyret.org/pull/244/files

jpolitz:
Probably the spec here is to create a new chart type called `residual-chart` that takes a list of xs, a list of ys, and a list of residuals.

It renders vertical bars from the y locations in correct proportion for the magnitude of the residual. (The black bars in @schanzer 's picture)

schanzer:
Adding notes from my call with @ds26gte 

- Google Charts _already has an [interval chart](https://developers.google.com/chart/interactive/docs/gallery/intervals)_, which can be configured to draw intervals as lines, error bars, area charts, and more. 
[...]

PR #489, opened 2023-10-18
Adds residual chart (issue #486)

ds26gte:
This adds the function residual-chart-from-list that takes xs, ys, and residuals and draws a line plot with stick intervals from each y to its corresponding residual.

### DIFF
--- a/src/web/arr/trove/chart.arr
+++ b/src/web/arr/trove/chart.arr
@@ -1039,6 +1039,27 @@ default-bar-chart-series = {
   default-interval-color: none
 }
 
+type IntervalChartSeries = {
+  tab :: TableIntern,
+  axisdata :: Option<AxisData>,
+  color :: Option<I.Color>,
+  colors :: Option<RawArray<I.Color>>,
+  pointers :: Option<RawArray<Pointer>>,
+  pointer-color :: Option<I.Color>,
+  horizontal :: Boolean,
+  default-interval-color :: Option<I.Color>
+}
+
+default-interval-chart-series = {
+  color: none,
+  colors: none,
+  pointers: none,
+  pointer-color: none,
+  axisdata: none,
+  horizontal: false,
+  default-interval-color: none,
+}
+
 type MultiBarChartSeries = { 
   tab :: TableIntern,
   axisdata :: Option<AxisData>,
@@ -1235,6 +1256,27 @@ default-bar-chart-window-object :: BarChartWindowObject = default-chart-window-o
   y-max: none,
 }
 
+type IntervalChartWindowObject = {
+  title :: String,
+  width :: Number,
+  height :: Number,
+  backgroundColor :: Option<I.Color>,
+  borderSize :: Number,
+  borderColor :: Option<I.Color>,
+  render :: ( -> IM.Image),
+  x-axis :: String,
+  y-axis :: String,
+  y-min :: Option<Number>,
+  y-max :: Option<Number>,
+}
+
+default-interval-chart-window-object :: IntervalChartWindowObject = default-chart-window-object.{
+  x-axis: '',
+  y-axis: '',
+  y-min: none,
+  y-max: none,
+}
+
 type HistogramChartWindowObject = {
   title :: String,
   width :: Number,
@@ -1378,6 +1420,26 @@ data DataSeries:
     error-bars: single-error-bars-method,
     interval-color: interval-color-method, 
     constr: {(): bar-chart-series},
+  | interval-chart-series(obj :: IntervalChartSeries) with:
+    is-single: false,
+    color: color-method,
+    colors: color-list-method,
+    sort: default-sort-method,
+    sort-by: sort-method,
+    sort-by-label: label-sort-method,
+    add-pointers: axis-pointer-method,
+    pointer-color: pointer-color-method,
+    format-axis: format-axis-data-method,
+    make-axis: make-axis-data-method,
+    scale: scale-method,
+    method horizontal(self, b :: Boolean):
+      self.constr()(self.obj.{horizontal: b})
+    end,
+    annotations: single-annotations-method,
+    intervals: single-intervals-method,
+    error-bars: single-error-bars-method,
+    interval-color: interval-color-method,
+    constr: {(): interval-chart-series},
   | multi-bar-chart-series(obj :: MultiBarChartSeries) with: 
     is-single: true,
     colors: color-list-method,
@@ -1459,6 +1521,12 @@ data ChartWindow:
     max: max-method,
   | bar-chart-window(obj :: BarChartWindowObject) with:
     constr: {(): bar-chart-window},
+    x-axis: x-axis-method,
+    y-axis: y-axis-method,
+    y-min: y-min-method,
+    y-max: y-max-method,
+  | interval-chart-window(obj :: IntervalChartWindowObject) with:
+    constr: {(): interval-chart-window},
     x-axis: x-axis-method,
     y-axis: y-axis-method,
     y-min: y-min-method,
@@ -1873,6 +1941,38 @@ fun stacked-bar-chart-from-list(
   data-series.make-axis(max-positive-height, max-negative-height)
 end
 
+fun residual-chart-from-list(
+  xs :: P.LoN,
+  ys :: P.LoN,
+  residuals :: P.LoN
+) -> DataSeries block:
+  doc: ```
+       Consumes a list of x's, a list of y's, and a list of residuals
+       and constructs an line plot with stick intervals pointing
+       from each y to the corresponding residual
+       ```
+  xs-length = xs.length()
+  ys-length = ys.length()
+  residuals-length = residuals.length()
+  when xs-length <> ys-length:
+    raise('residual-chart: xs and ys should have the same length')
+  end
+  when xs-length <> residuals-length:
+    raise('residual-chart: residuals should have the same length as xs and ys')
+  end
+  when xs-length == 0:
+    raise('residual-chart: need at least one datum')
+  end
+  xs.each(check-num)
+  ys.each(check-num)
+  residuals.each(check-num)
+
+  residual-strings = map(num-to-string, residuals)
+  default-interval-chart-series.{
+    tab: to-table3-n(xs, ys, residuals), #new
+  } ^ interval-chart-series
+end
+
 fun box-plot-from-list(values :: P.LoLoN) -> DataSeries:
   doc: "Consume values, a list of list of numbers and construct a box chart"
   labels = for map_n(i from 1, _ from values): [sprintf: 'Box ', i] end
@@ -2026,6 +2126,13 @@ fun render-chart(s :: DataSeries) -> ChartWindow:
           P.bar-chart(self, obj)
         end
       } ^ bar-chart-window
+    | interval-chart-series(obj) =>
+      default-interval-chart-window-object.{
+        method render(self):
+          _ = check-render-y-axis(self)
+          P.interval-chart(self, obj)
+        end
+      } ^ interval-chart-window
     | multi-bar-chart-series(obj) => 
       default-bar-chart-window-object.{
         method render(self):
@@ -2449,4 +2556,5 @@ from-list = {
   freq-bar-chart: freq-bar-chart-from-list,
   labeled-box-plot: labeled-box-plot-from-list,
   box-plot: box-plot-from-list,
+  residual-chart: residual-chart-from-list,
 }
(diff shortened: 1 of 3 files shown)


schanzer:
@ds26gte this is really exciting! I tried the following program, and the resulting chart isn't quite what I expected.
```
include chart
xs  = [list: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
ys  = [list: 2, 4, 6, 8,10,12,14,16,18, 20]
res = [list: 1, 2, 4,-1,-2,-4, 0, 4,-3,  3]

series = from-list.residual-chart(xs, ys, res)
c = render-chart(series)
c.display()
```
image

Some of those residuals should be on the other side of the line. Am I doing something wrong here?

src/web/js/trove/chart-lib.js
```diff
@@ -771,6 +772,35 @@
...
+    data.addColumn('number', 'x');
+    data.addColumn('number', 'values');
+    data.addColumn({id: 'i0', type: 'number', role: 'interval'});
+    data.addColumn({id: 'i1', type: 'number', role: 'interval'});
+
+    data.addRows(table.map(row => [row[0], row[1], row[1], row[2]]));
+
+    const options = {
+      curveType: 'function',
+      lineWidth: 2,
```
schanzer: This should be 0, since we ONLY want to see the residuals.

shriram:
@schanzer Nothing wrong, I just think you've forgotten about a little thing called gravity.

shriram:
@schanzer Nothing wrong, I just think you've forgotten about a little thing called gravity.

schanzer:
Ah yes! The residuals are the y', not delta-y. Silly me 🤦‍♂️

schanzer:
@jpolitz I'll defer to you on the design of the code, but this raises no red flags for me. 

But in reading through this, I think we need a lower-level Pyret-facing API that includes some configurability so that we can use intervals in other areas as well. From reading [the documentation](https://developers.google.com/chart/interactive/docs/gallery/intervals), it looks like we want interval-chart to take in a list of xs, ys, a _list of lists_ (each of which represents an interval). I propose that it also expose a method `.type()`, which allows for stick, bar, box, and point intervals. I could also see a universe in which it exposes a `.color()` method, which colors the bars/sticks/boxes/points.

This would allow us to replicate most of the interval types in Google's API, though only one-at-a-time. But that's easily addressed by composing multiple interval charts view `render-charts`, so ultimately this would give us quite a bit of power.

The functionality of `residual-chart` could still exist, being built on top of this either as part of the chart library or implemented via a Bootstrap library.

ds26gte:
> Ah yes! The residuals are the y', not delta-y. Silly me 🤦‍♂️

Corrected. _Residual_ does mean what you expected, so I've corrected the 3rd arg to be _Δy_ rather than _y'_. 

(Of course, please check that the direction of the offset is also as you would expect. I really don't know for sure.)

src/web/arr/trove/chart.arr
```diff
@@ -1046,6 +1046,8 @@ type IntervalChartSeries = {
   colors :: Option<RawArray<I.Color>>,
   pointers :: Option<RawArray<Pointer>>,
   pointer-color :: Option<I.Color>,
+  point-size :: Number,
+  lineWidth :: Number,
```
schanzer, line 1056: For consistency's sake, shouldn't this be `line-width`?

src/web/arr/trove/chart.arr
```diff
@@ -1055,6 +1057,8 @@ default-interval-chart-series = {
   colors: none,
   pointers: none,
   pointer-color: some(C.color(228, 147, 7, 1)),
+  point-size: 4,
+  lineWidth: 1,
```
schanzer, line 1067: `line-width`?

src/web/arr/trove/chart.arr
```diff
@@ -1039,6 +1039,31 @@ default-bar-chart-series = {
   default-interval-color: none
 }
 
+type IntervalChartSeries = {
+  tab :: TableIntern,
+  axisdata :: Option<AxisData>,
+  color :: Option<I.Color>,
+  colors :: Option<RawArray<I.Color>>,
```
schanzer: What's the plural `colors` for here? I don't think it's used in the JS implementation

src/web/arr/trove/chart.arr
```diff
@@ -1039,6 +1039,31 @@ default-bar-chart-series = {
...
+  pointer-color :: Option<I.Color>,
+  point-size :: Number,
+  lineWidth :: Number,
+  horizontal :: Boolean,
+  default-interval-color :: Option<I.Color>
+}
+
+default-interval-chart-series = {
+  color: some(C.red),
+  colors: none,
```
schanzer: same question - what are `colors` for here?

src/web/arr/trove/chart.arr
```diff
@@ -1039,6 +1039,31 @@ default-bar-chart-series = {
...
+  pointers: none,
+  pointer-color: some(C.color(228, 147, 7, 1)),
+  point-size: 4,
+  lineWidth: 1,
+  axisdata: none,
```
schanzer, line 1068: should this be `axis-data`, in keeping with Pyret conventions?

src/web/arr/trove/chart.arr
```diff
@@ -1873,6 +1945,38 @@ fun stacked-bar-chart-from-list(
   data-series.make-axis(max-positive-height, max-negative-height)
 end
 
+fun residual-chart-from-list(
```
schanzer: Can we rename this `interval-chart-from list`, in keeping with the data series renaming?

schanzer (changes requested):
@ds26gte left a few comments for you, but for the most part this looks _awesome!_

@jpolitz hoping you can weigh in on this as well.

src/web/arr/trove/chart.arr
ds26gte: Unfortunately, the existing series use camelCase and kebab-case for different fields, and I was trying to be consistent with them.

src/web/arr/trove/chart.arr
ds26gte: In case we use multiple residuals?

src/web/arr/trove/chart.arr
schanzer: makes sense. @jpolitz any interest in Dorai opening another PR in which these get normalized across the charting library?

src/web/arr/trove/chart.arr
schanzer: Multiple residuals should be handled via multiple instances of `interval-chart`, composed via `render-charts(...)`

src/web/arr/trove/chart.arr
```diff
@@ -334,7 +334,7 @@ line-width-method = method(self, lineWidth :: Number) block:
 end
 
 style-method = method(self, style :: String) block:
-  when not(string-equal(style, "sticks")) and not(string-equal(style, "bars")):
+  when not(string-equal(style, "sticks")) and not(string-equal(style, "bars")) and not(string-equal(style, "boxes")):
     raise("style: must be either sticks or bars")
```
schanzer: @ds26gte shouldn't this error include boxes as a recommended type?

src/web/arr/trove/chart.arr
```diff
@@ -1873,6 +1945,38 @@ fun stacked-bar-chart-from-list(
   data-series.make-axis(max-positive-height, max-negative-height)
 end
 
+fun residual-chart-from-list(
+  xs :: P.LoN,
+  ys :: P.LoN,
+  residuals :: P.LoN
```
schanzer: This shouldn't be called residuals now that we've renamed the series.

src/web/arr/trove/chart.arr
```diff
@@ -1873,6 +1945,38 @@ fun stacked-bar-chart-from-list(
...
+) -> DataSeries block:
+  doc: ```
+       Consumes a list of x's, a list of y's, and a list of residuals
+       and constructs an line plot with stick intervals pointing
+       from each y to the corresponding residual
```
schanzer: We should remove the "residual" text, and update accordingly

src/web/arr/trove/chart.arr
```diff
@@ -1873,6 +1945,38 @@ fun stacked-bar-chart-from-list(
...
+       ```
+  xs-length = xs.length()
+  ys-length = ys.length()
+  residuals-length = residuals.length()
```
schanzer: Rename

src/web/arr/trove/chart.arr
```diff
@@ -1873,6 +1945,38 @@ fun stacked-bar-chart-from-list(
...
+  when xs-length <> ys-length:
+    raise('residual-chart: xs and ys should have the same length')
```
schanzer: Rename

src/web/arr/trove/chart.arr
```diff
@@ -1873,6 +1945,38 @@ fun stacked-bar-chart-from-list(
...
+  end
+  when xs-length <> residuals-length:
+    raise('residual-chart: residuals should have the same length as xs and ys')
```
schanzer: rename

src/web/arr/trove/chart.arr
```diff
@@ -1873,6 +1945,38 @@ fun stacked-bar-chart-from-list(
...
+  end
+  when xs-length == 0:
+    raise('residual-chart: need at least one datum')
```
schanzer: rename

src/web/arr/trove/chart.arr
```diff
@@ -1873,6 +1945,38 @@ fun stacked-bar-chart-from-list(
...
+  end
+  xs.each(check-num)
+  ys.each(check-num)
+  residuals.each(check-num)
```
schanzer: rename

src/web/arr/trove/chart.arr
```diff
@@ -1873,6 +1945,38 @@ fun stacked-bar-chart-from-list(
...
+  yprimes = map2(lam(y, offset): y + offset end, ys, residuals)
```
schanzer: rename

src/web/arr/trove/chart.arr
```diff
@@ -2449,4 +2560,5 @@ from-list = {
   freq-bar-chart: freq-bar-chart-from-list,
   labeled-box-plot: labeled-box-plot-from-list,
   box-plot: box-plot-from-list,
+  residual-chart: residual-chart-from-list,
```
schanzer: I think you've already renamed this -- now I'm wondering if Git is just showing me an old version of the file?

schanzer (changes requested):
Minor nits - I think this is good to go once the `residual` naming is gone.

src/web/arr/trove/chart.arr
jpolitz: No issue with it, though having it separate is preferred.

jpolitz:
Want to make sure I understand the expected behavior of `interval-chart-from-list`. Here it's drawing the lines only below the dot. Also, the really small delta seems to cause the whole point to not be rendered?

image

src/web/js/trove/chart-lib.js
```diff
@@ -771,6 +772,59 @@
...
+
+    const stickColor = get_default_color(rawData);
+    const stickWidth = get(rawData, 'lineWidth');
+
+    const pointColor = get_pointer_color(rawData);
+    const pointSize = get(rawData, 'point-size');
+
+    const fillOpacity = (style == 'boxes') ? 0 : 1;
+
+    data.addRows(table.map(row => [row[0], row[1], row[1], row[2]]));
```
jpolitz: It's likely that the data should not be passed through uninterpreted. Probably it would be correct to call `toFixnum` on all of these if they are all numbers (and matching the columns above). e.g. https://github.com/brownplt/code.pyret.org/blob/08ad7d3e876030e0013f8b889d0d3e9d2d99d4ea/src/web/js/trove/chart-lib.js#L504

Otherwise, roughnums will cause issues.

src/web/js/trove/chart-lib.js
```diff
@@ -771,6 +772,59 @@
...
+
+    const options = {
+      curveType: 'function',
+      lineWidth: 0,
+      intervals: { style:'sticks', lineWidth: 2,  },
+      interval: {
+        'i0': {
+          'style': style,
+          'color': stickColor,
+          'lineWidth': stickWidth,
```
jpolitz, line 808: Similarly the widths and such, if they could come in as roughnums from Pyret, need to be converted somehow.

src/web/arr/trove/chart.arr
jpolitz: (Separate PR I mean, not separate naming conventions :-))

jpolitz:
Great. Thanks for the tests!

PR #497, opened 2023-12-14
- Allow interval charts to compose w/other charts that render-charts() takes

ds26gte:
#486
- chart-lib.js: Have plot() recognize and process interval charts, instead of having separate intervalChart()
- chart.arr: Rewrite render-charts() to accept interval-chart-series. For this, make interval-chart-series closer in properties to function/line/scatter series. interval-chart-from-list() now must keep track of the largest y's (including offset) so render-charts() can create a better bounding box. render-chart() no longer needs to do anything special with interval-charts, passing them on to render-charts()

### DIFF
--- a/src/web/js/trove/chart-lib.js
+++ b/src/web/js/trove/chart-lib.js
@@ -1135,11 +1135,11 @@
     const scatters = get(rawData, 'scatters');
     const lines = get(rawData, 'lines');
     const intervals = get(rawData, 'intervals');
+    const minIntervalIndex = scatters.length + lines.length;
     const data = new google.visualization.DataTable();
     data.addColumn('number', 'X');
-    const combined = scatters.concat(lines);
-    console.log('combined is ', JSON.stringify(combined));
-    const combined2 = combined.concat(intervals);
+    const combined = scatters.concat(lines).concat(intervals);
+    console.log('combined is', combined);
     const legends = [];
     let cnt = 1;
     const legendEnabled = combined.length > 1;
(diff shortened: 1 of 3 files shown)


schanzer:
@ds26gte Here's the sample program I'm using to test, which combines a scatter plot and an interval chart:
```
include chart

xs = [list: 1,-2,9,4]
ys = [list: 1,2,3,4]
fn = lam(x): x end
residuals = map2(
  lam(y, prediction): y - prediction end, 
  ys, 
  map(fn, xs))


scatter = from-list.scatter-plot(xs, ys)
fn-plot = from-list.function-plot(fn)
intervals = from-list.interval-chart(xs, ys, residuals)

render-charts([list: scatter, fn-plot, intervals])
  .display()
```
With this PR, I get the following error:
image

jpolitz:
Is this at a point where it should be good to review + go or still in progress?

schanzer:
@jpolitz good to go! Ready for final review and merger

src/web/js/trove/chart-lib.js
```diff
@@ -1207,36 +1152,66 @@
...
     });
 
+    // console.log('legends is', legends);
+    // console.log('data I is', data);
+
     combined.forEach((p, i) => {
-      /*
-      x | n n n | y | n n n n n n n n n n n n
-            i         combined.length - i - 1
-      */
```
jpolitz, line 1216: This comment was informative about the shape of the data – did the shape change/is the comment irrelevant? Was it wrong/should it be documented somewhere what the new shape is?

src/web/js/trove/chart-lib.js
```diff
@@ -1207,36 +1152,66 @@
...
+              ${labelRow}`;
+            currentRow[4*i + 3] = row[1];
+            currentRow[4*i + 4] = row[2];
+          }
+          return currentRow;
+        }));
+      };
     });
 
+    // console.log('row setting done');
```
jpolitz, line 1213: Lots of commented out `console.log`s – typically our style is to remove these though if there's a good reason for leaving them in because they were difficult to design could leave a comment to that effect.

jpolitz:
@ds26gte @schanzer I think for this go ahead and push directly to `horizon` so you can get the testing and autodeploy. Might make testing cycles faster! I won't release till you say it's ready.

ds26gte:
@schanzer brought up a new test case which I'll investigate and then let you know after if it's ready to go.

PR #504, opened 2024-01-21
Enter submits modals.

### LINKED ISSUE #277: Pressing Enter in rename/save a copy dialog, etc, should trigger saving
It's annoying to have to use the mouse/tab to get to the Submit button.

asolove:
Resolves #277.

We already auto-focus the submit button for modals with no inputs, so enter already works there. With this change, if the modal has text inputs, hitting enter while focused on one of these text inputs will also immediately try to submit.

If we ever add modals that have inputs where enter should do something else (like a multi-line textarea comment box), we'll need to ensure that input captures and does not bubble the enter event, so that it doesn't trigger this.

### DIFF
--- a/src/web/js/modal-prompt.js
+++ b/src/web/js/modal-prompt.js
@@ -111,6 +111,12 @@ define("cpo/modal-prompt", ["q"], function(Q) {
       this.submitButton.show();
     }
     this.closeButton.click(this.onClose.bind(this));
+    this.modal.keypress(function(e) {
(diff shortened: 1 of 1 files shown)


blerner:
This code as-is obviously works :)  But for future robustness -- would it be easier to add an option `enterSubmits: boolean` that decides whether to add this submit-on-press-Enter behavior?

ds26gte:
reviewed w/Joe

PR #509, opened 2024-02-10
Improve styling of Rename modal and modal cancel buttons.

### LINKED ISSUE #350: better labels

asolove:
Resolves #350.

| Before | After |
| ---- | ---- |
| image | image |

- Rename modal's submit button now says Rename
- All modal cancel buttons have blue, rather than gray, text, to show they are clickable and to match Google styles.

In the initial issue, @shriram asked:

> I also feel like the very light grey of Cancel makes it looks like it's not enabled. Is the grey instead of red for cancel a consistent style?

I agree that the light gray only makes it look not enabled. This PR makes the text blue to match the styles commonly used in the Google/Chromebook apps that students will be familiar with. e.g. here is a sample from Google Docs:

image

In general, I'd recommend we not use red for non-destructive secondary actions like "close", as it has a connotation of something dangerous, whereas in this case close should clearly be taken as the neutral action with no side-effects. CPO no longer uses red for any secondary actions that I can find from clicking around.

### DIFF
--- a/src/web/css/editor.css
+++ b/src/web/css/editor.css
@@ -2379,7 +2379,7 @@ table.pyret-row {
 .modal .close{
   border: 1px solid #ccc;
   background: white;
-  color: #999;
+  color: var(--run-bg);
   margin-right: 16px;
 }
 
--- a/src/web/js/beforeBlocks.js
+++ b/src/web/js/beforeBlocks.js
@@ -795,6 +795,7 @@ $(function() {
         title: "Rename this file",
         style: "text",
         narrow: true,
+        submitText: "Rename",
         options: [
           {
             message: "The new name for the file:",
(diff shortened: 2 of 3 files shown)


shriram:
Oh yeah, I really like this improvement! And matching GDocs seems to have
value on Chromebooks and elsewhere. Thanks!

jpolitz:
Looks good thanks!

Issue #501, opened 2024-01-17
Pyret Blocks (tracking issue)


schanzer:
Now that the pcardune-blocks branch has been merged into horizon, we need a tracking issue to keep tabs on the remaining work.

- [ ] include and import need to work correctly, with snap's string-handling being a bit of an obstacle
- [ ] do we want to support dragging a block into the interactions area?
- [ ] we need to be able to share source locations (from error messages) with the block highlighter

asolove:
For item 3, highlighting error locations, see: https://github.com/brownplt/code.pyret.org/pull/502

asolove:
A couple other topics:

- The current Pyret blocks are mostly expression-oriented, while Pyret has for loops and multi-line statement contexts that don't seem to be supported yet in blocks. I assume there's some work on our side to tell Snap which holes support that and to add more block types?

- Hiding Pyret syntax: the blocks environment ends up not looking a lot like Pyret syntax (there's no `end`, quotation around strings is omitted, etc.). So it seems like there are a few options:
  - accept such differences and have teachers learn full Pyret so they can help students?
  - keep the blocks syntax as-is, and migrate anything that shows "normal" Pyret to instead show the blocks syntax (docs, instructional material and error messages)
  - try to make the blocks syntax more like Pyret (with explicit quotes, "end", etc.).

shriram:
As an aside, note that Pyret `for` loops are not actually statements! They're genuine expressions. E.g.:
```
squares = for map(i from range(0, 5)): i * i end
check:
  squares is [list: 0, 1, 2 * 2, 3 * 3, 4 * 4]
end
```
In Pyret, `for` is just a very fancy and roundabout way of writing `lambda`, sorry-not-sorry. (-:

asolove:
Of course, and thank you for the callout as I was being very imprecise with my language.

What I mean to say is not about semantics but instead:
- lexical syntax: what are we going to do about line breaks and indentation?
- syntax: will we allow multiple statements inside a single hole?

**Lexical syntax**
As expressions get longer, Blocks world shows them on a single line until it runs out of space, at which point it breaks the line and indents just enough to leave the expression inside the circle of its parent:
<img width="335" alt="image" src="https://github.com/brownplt/code.pyret.org/assets/8495/85c56d3c-c4a5-4a30-abd1-fef3841397b7">


The question here is whether this automatic line breaking is sufficient and understandable to the intended audience as you make more and more complex nested expressions for things like the flag assignment.

[...]

PR #513, opened 2024-03-05
Blocks language supports benchmark program

asolove:
This PR contains changes to the Snap definitions of the Pyret language to enable us to construct the demo program we want to be able to run [the benchmark program](https://code.pyret.org/editor#share=1hmVFpA2zWBWaKulm6LwHHzKrDhH15jvP).

In [previous testing](https://github.com/brownplt/code.pyret.org/issues/501#issuecomment-1970138201), we found a few missing items:

- [x] Introduce a new variable and assign it to a value: `var = val`
- [x] Lookup for `Row`, `StringDict` and other map-like things: `map[key]`
    image

- [x] image-scatter-plot (turns out this already worked and Adam just misunderstood)
- [x] load-table can define multiple column names as variables
    image

- [x] allow multiple arguments in function definitions and applications
    image

### DIFF
--- a/src/web/js/transpile.xml
+++ b/src/web/js/transpile.xml
@@ -80,7 +80,23 @@ else:
                     <translations />
                     <inputs>
                         <input type="%upvar">name</input>
-                        <input type="%n">val</input>
+                        <input type="%s">val</input>
+                    </inputs>
+                    <script>
+                        <block s="doReport">
+                            <l />
+                        </block>
+                    </script>
+                </block-definition>
+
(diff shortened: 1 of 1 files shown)


asolove:
@jpolitz I think this is everything needed to run our benchmark program. There are only changes to the blocks definitions. It would be helpful if we could merge this to horizon and deploy. Then we can test it out in an environment that can `use context shared-gdrive`.

jpolitz:
I believe this is now live at https://pyret-horizon.herokuapp.com/blocks

PR #515, opened 2024-03-12
Check that color rgba values are within ranges at construction time.

### LINKED ISSUE #487: `color` does not enforce refinements
The Color library includes the `color` constructor, whose documentation says the first three inputs should all be numbers between 0-255 (inclusive). However, calling the function with numbers outside of that range seems to work just fine.

This is either a bug in the implementation or the documentation. I'm inclined to say the implementation should conform to the documentation.

asolove:
Resolves #487.

The author has checked that this works in basic cases, but isn't sure this is safe all the time. (Are there programs that currently rely on the clamping error, that this will cause to error?)

### DIFF
--- a/src/web/arr/trove/color.arr
+++ b/src/web/arr/trove/color.arr
@@ -6,12 +6,20 @@ provide-types *
 import global as _
 import base as _
 
+fun zero-to-255(n :: Number) -> Boolean:
+  (n >= 0) and (n <= 255)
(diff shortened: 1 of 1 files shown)


asolove:
See discussion for why this is a bad idea: https://github.com/brownplt/code.pyret.org/issues/487#issuecomment-1992774081

PR #516, opened 2024-03-13
Choose Context modal doesn't show bad non-context strings

### LINKED ISSUE #399: Choosing context on a file with no context line substrings into whatever is there
This won't come up that often, but when it does, it's weird:

image

image

Need to be a *little* smarter here. (Underlying mistake was implementing the dialog assuming we would add context lines to existing files, then realizing that was a bad idea)

asolove:
Fixes #399

Previously, if your first line wasn't a `use context` declaration, the Choose Context modal would just grab an arbitrary substring of your first line, resulting in gibberish.

This change first checks if the first line is actually a `use context`. If so, it shows the current context. If not, it shows the current context as blank and lets you enter a new one.

### DIFF
--- a/src/web/js/beforePyret.js
+++ b/src/web/js/beforePyret.js
@@ -447,6 +447,8 @@ $(function() {
     $("#download").append(downloadElt);
   });
 
+  const CONTEXT_PREFIX = "use context ";
+
   function showModal(currentContext) {
     function drawElement(input) {
       const element = $("<div>");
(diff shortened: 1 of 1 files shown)


src/web/js/beforePyret.js
```diff
@@ -447,6 +447,8 @@ $(function() {
     $("#download").append(downloadElt);
   });
 
+  const CONTEXT_PREFIX = "use context ";
```
blerner, line 450: This should be `/^use context\s+/`, so that the `.slice` on line 498 trims any interstitial whitespace between the `context` and the actual context value.

Actually, could you reuse `firstLineIsNamespace` from `beforePyret`, and expose it along with `setContextLine`, at https://github.com/brownplt/code.pyret.org/blob/ef0d4daa5c52ca28c09d659fde5b4868b95758f3/src/web/js/beforePyret.js#L294, so that you can just do `CPO.editor.firstLineIsNamespace`?  You'd fix up the regex on line 225, (it shouldn't end in `.*`, but rather `\s+`.  Then maybe refactor lines 495-498 of your code below into a function `getContextFromFirstLine()` that is likewise made available?

blerner (commented):
LGTM, just some minor refactoring to make it less repetitive with existing code.

PR #519, opened 2024-04-04
upgrade and fix pie chart rendering

schanzer:
[Based on a bug report from a teacher](https://discourse.bootstrapworld.org/t/image-pie-chart/2738/1)

It turns out that unlike scatter plots, histograms, and all manner of bar charts, _the order of SVG pie slices does not follow the order of the rows in the table!_ This PR makes the following changes:

- Compute the correct ordering of SVG slices before drawing the new ones
- Draw the new slices at a better location in the DOM, which fixes a number of display issues regarding interactive labels
- Draw tiny versions of the images in the legend as well


src/web/js/trove/chart-lib.js
```diff
@@ -475,6 +475,7 @@
 
   function pieChart(globalOptions, rawData) {
     const table = get(rawData, 'tab');
+    console.log(table)
```
sorawee: A leftover debugging?

src/web/js/trove/chart-lib.js
```diff
@@ -530,21 +532,34 @@
           // The only way to hijack marker events is to walk the DOM here
           // If Google changes the DOM, these lines will likely break
           const svgRoot = chart.container.querySelector('svg');
-          const slices = [...svgRoot.children].slice(2, -1);
+
+          // The order of SVG slices is *not* the order of the rows in the table!!
+          // 2 slices: drawn in reverse order
+          // >2 slices: the first row in the table is the
```
sorawee: Sounds like the comment is incomplete.

schanzer:
Thanks, @sorawee ! I've pushed fixes for these.

src/web/js/trove/chart-lib.js
sorawee: @schanzer I meant "is the" seems to abruptly end. But it's great that this is now more exhaustive as well!

src/web/js/trove/chart-lib.js
```diff
@@ -530,21 +531,34 @@
...
           // If Google changes the DOM, these lines will likely break
           const svgRoot = chart.container.querySelector('svg');
-          const slices = [...svgRoot.children].slice(2, -1);
+
+          // The order of SVG slices is *not* the order of the rows in the table!!
+          // 1 or 2 slices: drawn in reverse order
+          // >2 slices: the first row in the table is the 
+          let slices;
+          if(table.length <= 2) {
+            slices = [...svgRoot.children].slice(2, -1).reverse();
```
blerner: More efficient to call `Array.prototype.slice.call(svgRoot.children, 2, -1)`, and avoid the extra `[...svgRoot.children]` array creation.

src/web/js/trove/chart-lib.js
schanzer: Oh! I didn't even noticed that got cut off. Fixed in the latest commit

src/web/js/trove/chart-lib.js
schanzer: good call! fixed.

schanzer:
@blerner I've added tests (all of which pass when run in the browser), but I wasn't able to run `npm run test` from my command line due to missing packages.

blerner:
I'm going to assume that this PR does what @schanzer needs it to do, and that the remaining branch CI failure is because this branch doesn't yet have @jpolitz 's fix from #horizon.

PR #527, opened 2024-05-01
bulk npm update

blerner:
If this passes tests, we should just merge this and close out all the stale dependabot PRs...


jpolitz:
Nice, thanks.

PR #533, opened 2024-05-02
Add string-find-index and string-get-index

blerner:
See https://github.com/brownplt/pyret-lang/pull/1739


blerner:
Note: this PR cannot land until after https://github.com/brownplt/pyret-lang/pull/1739 lands, so the build-breakage right now is not a problem.

PR #535, opened 2024-05-03
An initial attempt to fix #534

### LINKED ISSUE #534: rendering tables
Is there any chance of getting a useful `torepr`/`tostring` for tables?

Tables are a pretty useful data structure for reactor worlds, and having a default rendering means it's a lot easier to view/debug what is going on without having to create a complex rendering operation. Right now they just produce `<table>`.

blerner:
The request is to render the `display`ed version of a Pyret value instead of its `to-repr` rendering.  Note that this interacts oddly with mouse handlers and key handlers, since displayed values are interactive but so is a reactor, and keyboard focus can get screwed up.  There's also no limit to the size of the window; we might want to impose a maximum size and a overflow: scroll on these...but I don't understand the cssFunction well enough to know how to apply that.

Potentially fixes #534

### DIFF
--- a/src/web/js/trove/world.js
+++ b/src/web/js/trove/world.js
@@ -533,13 +533,13 @@
     DefaultDrawingOutput.prototype.toRawHandler = function(toplevelNode) {
       var that = this;
       var worldFunction = function(world, success) {
-        var textNode = jQuery("<pre>");
+        var hostNode = document.createElement("div");
         return runtime.safeCall(function() {
-          return runtime.toReprJS(world, runtime.ReprMethods._torepr);
-        }, function(str) {
-          textNode.text(str);
+          return runtime.toReprJS(world, runtime.ReprMethods['$cpo']);
+        }, function(jNode) {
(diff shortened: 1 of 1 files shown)


blerner:
Here's a screenshot of a silly example.  Note that you cannot get interactivity with the Pyret values in here *unless* there is no `on-mouse` handler, since that'll steal mouse focus first.
```
r = reactor:
  init: table: n, o, p end,
  on-key: 
    {(t, _):
      t.add-row(t.row({ x: random(100), y: random(200) }, { x: random(100), y: random(200) }, { x: random(100), y: random(200) }))
    },
end

r.interact()
```
image

blerner:
@shriram ^

@jpolitz please review; I _think_ this is safe since the original `toRepr` call was already in a `safeCall`, so this _should_ be stack-safe, but I'm not sure what other interactions might break because of this.

jpolitz:
Huh, cool. I can't see this being worse than what we have.

PR #550, opened 2024-08-29
@kfisler's dcic2024.arr context file added

ds26gte:
Kathi supplied this context file for use with DCIC.

### DIFF
--- a/src/web/arr/trove/dcic2024.arr
+++ b/src/web/arr/trove/dcic2024.arr
@@ -1,3 +1,4 @@
+use context starter2024
 ################################################################
 # CS111/DCIC Support functions
 # August 2024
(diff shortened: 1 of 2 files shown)


blerner:
The build failed due to unbound identifiers...

ds26gte:
> The build failed due to unbound identifiers...

FIxed by explicitly using context starter2024. OK to merge?

blerner:
This is likely not quite what you mean to do -- right now this context _only_ provides 26 table functions, and nothing else.  Pyret doesn't implicitly re-provide imported things (this is analogous to `require` and `provide` in Racket). If you're trying to build atop `starter2024`, then you probably need to

```
use context empty
import starter2024 as S
provide from S: *, type *, data * end
include from S: *, type *, data * end

...rest as written...
```
You also have no contracts on any of the functions (they're in comments only), so you'll get no annotation checking on them at all.

I'd recommend trying to use this with some DCIC programs to see if it actually works for them as intended, and you'll pretty quickly find the missing exports.

PR #551, opened 2024-09-14
add support for vim in the editor

g-w1:


src/web/js/beforePyret.js
```diff
@@ -188,20 +188,24 @@ $(function() {
...
-        "Esc Left": "goBackwardSexp",
         "Alt-Left": "goBackwardSexp",
-        "Esc Right": "goForwardSexp",
         "Alt-Right": "goForwardSexp",
         "Ctrl-Left": "goBackwardToken",
         "Ctrl-Right": "goForwardToken",
         [`${modifier}-/`]: "toggleComment",
-      }),
+    });
+    CPO.noVimKeyMap = CodeMirror.normalizeKeyMap({
```
g-w1, line 202: Please let me know if there is a better place to store the non-vim specific keymap than in `CPO` :)

PR #559, opened 2024-11-13
Num dot charts

ds26gte:
Adds `num-dot-chart-from-list` for numerical (in contrast to categorical) dot charts.  It takes a list of numbers, which may repeat and need not be ordered, and creates a dot chart that responds to chart resizing by piling up the dots vertically when there isn't enough horizontal elbow room for different dots not to overlap. Even when a dot climbs up vertically, its x-position remains true. Also, if a dot's x-value precedes another in the input, then when one ends up piled atop the other, the left dot is lower than the right dot.

### DIFF
--- a/src/web/js/trove/chart-lib.js
+++ b/src/web/js/trove/chart-lib.js
@@ -1624,20 +1624,23 @@
             // const circleR = toFixnum(get(combined[0], 'point-size'));
             const circleR = Number(circle0.getAttribute('r'));
             const offsetQuantum = 2 * circleR;
-            let lastX = false;
-            let currentNumOffsets = 0;
+            let prevDotArray = [];
+            function tooClose(x, y) {
+              return prevDotArray.some(function(n) {
+                return ((Math.abs(x - n[0]) < offsetQuantum) &&
+                  (Math.abs(y - n[1]) < offsetQuantum));
+              });
+            }
+
             circles.forEach((circle) => {
               // console.log('updating circle', i);
               const circleX = Number(circle.getAttribute('cx'));
-              if (lastX &&
-                (lastX === circleX || (circleX - lastX) < offsetQuantum)) {
-                // circle.setAttribute('cx', lastX); // if we don't want horiz stagger
-                currentNumOffsets++;
-              } else {
-                lastX = circleX; currentNumOffsets = 0;
+              let circleY = Number(circle.getAttribute('cy')) - 0.5*offsetQuantum;
+              while (tooClose(circleX, circleY)) {
+                circleY -= offsetQuantum;
               }
-              let circleY = Number(circle.getAttribute('cy'));
-              circleY -= (currentNumOffsets + 0.5) * offsetQuantum;
+
+              prevDotArray.push([circleX, circleY]);
 
               const circleElt = circle.cloneNode(false);
               circleElt.classList.add('__img_labels'); // tag for later gc
--- a/src/web/js/trove/chart-lib.js
+++ b/src/web/js/trove/chart-lib.js
@@ -1631,14 +1631,22 @@
               const circleX = Number(circle.getAttribute('cx'));
               if (lastX &&
                 (lastX === circleX || (circleX - lastX) < offsetQuantum)) {
-                circle.setAttribute('cx', lastX);
+                // circle.setAttribute('cx', lastX); // if we don't want horiz stagger
                 currentNumOffsets++;
               } else {
                 lastX = circleX; currentNumOffsets = 0;
               }
               let circleY = Number(circle.getAttribute('cy'));
               circleY -= (currentNumOffsets + 0.5) * offsetQuantum;
-              circle.setAttribute('cy', circleY);
+
+              const circleElt = circle.cloneNode(false);
+              circleElt.classList.add('__img_labels'); // tag for later gc
+              circleElt.setAttribute('cy', circleY);
+              circleElt.setAttribute('r', circleR);
+              circleElt.setAttribute('fill-opacity', 1);
+              Object.assign(circleElt, circle); // we should probably not steal *everything*...
+              circle.remove();
+              svgRoot.appendChild(circleElt);
             });
           }
         });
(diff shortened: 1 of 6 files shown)


src/web/js/trove/chart-lib.js
```diff
@@ -1253,10 +1254,16 @@
             } else {
               labelRow = '';
             }
+            let labelRowY = null;
```
schanzer, line 1257: I wouldn't overload this condition right now -- I would much rather you create an entirely separate condition for dot plots!

For now there will be a lot of duplicate logic (the first 5-10 LOC), but this can all be abstracted out once `ps` and `tab` are unified.

src/web/js/trove/chart-lib.js
```diff
@@ -1286,6 +1293,7 @@
 
     // ASSERT: if we're using custom images, *every* series will have idx 3 defined
     const hasImage = combined.every(p => get(p, 'ps').filter(p => p[3]).length > 0);
+    let dotChartP = false;
```
schanzer: curious about why you're not setting this directly, ie `const dotChartP = get(p, 'dot-chart')`

src/web/js/trove/chart-lib.js
```diff
@@ -1302,10 +1310,14 @@
         });
         // If we have our own image, make the point small and transparent
         if (i < scatters.length) {
+          if (get(p, 'dot-chart')) {
+            // console.log('dot chart scatter plot found');
+            dotChartP = true;
+          }
           $.extend(seriesOptions, {
-            pointSize: hasImage ? 1 : toFixnum(get(p, 'point-size')),
+            pointSize: (hasImage || dotChartP) ? 0.1 : toFixnum(get(p, 'point-size')),
```
schanzer: Rather than doing `A || B` everywhere, I'd recommend setting another predicate that's basically `replaceDefaultSVG`.

src/web/js/trove/chart-lib.js
```diff
@@ -1418,6 +1430,14 @@
       options['trendlines'][0]['degree'] = trendlineDegree;
     }
 
+    if (dotChartP) {
+      options['vAxis'] = {
+        ticks: [],
+        minValue: 0,
+        maxValue: 10,
```
schanzer, line 1438: We should comment this.

src/web/js/trove/chart-lib.js
```diff
@@ -1556,7 +1576,7 @@
             .append(redrawG);
         }
 
-        if(!hasImage) { return; } // If we don't have images, our work is done!
+        if(!hasImage && !dotChartP) { return; } // If we don't have images, our work is done!
```
schanzer: ...and this would become `if(!replaceDefaultSVG)`

src/web/js/trove/chart-lib.js
```diff
@@ -1601,6 +1622,43 @@
               svgRoot.appendChild(imageElt);
             });
           });
+          }
+
+          if (dotChartP) {
+            const circles = [...svgRoot.children[1].children[2].getElementsByTagName('circle')];
```
schanzer: Use `markers`, not SVG walking

src/web/js/trove/chart-lib.js
```diff
@@ -1601,6 +1622,43 @@
...
+                circleY -= offsetQuantum;
+              }
+
+              prevDotArray.push([circleX, circleY]);
+
+              const circleElt = circle.cloneNode(false);
+              circleElt.classList.add('__img_labels'); // tag for later gc
+              circleElt.setAttribute('cy', circleY);
+              circleElt.setAttribute('r', circleR);
+              circleElt.setAttribute('fill-opacity', 1);
```
schanzer, line 1658: We can drop `setAttribute` for `r` and `fill-opacity`.

schanzer (changes requested):

src/web/js/trove/chart-lib.js
ds26gte: `p` is not lexically available there. It is introduced by the function that's mapped across `combined`

src/web/js/trove/chart-lib.js
ds26gte: These seem to be needed. The circles/dots aren't drawn otherwise

src/web/js/trove/chart-lib.js
schanzer: @ds26gte your comment still doesn't address the important bit, which is the part you told me you "never would have guessed" yesterday: google charts needs some form of maxValue, otherwise having all the dots at y=0 will force the vAxis to be _centered_ at 0. 

If you never would have guessed this, I suspect it would be worth documenting in the comment!

src/web/js/trove/chart-lib.js
schanzer: Huh. Even if you use `cloneNode(true)`?

schanzer: Ah, of course -- because the circles we're cloning are already invisible. :P 

Nevermind. My bad.

src/web/js/trove/chart-lib.js
ds26gte: We are committed to eventually unifying the conditions for other reasons (i.e., getting rid of either `ps` or `tab`) so punting on this for now

src/web/js/trove/chart-lib.js
ds26gte: Done

src/web/js/trove/chart-lib.js
ds26gte: Done

src/web/js/trove/chart-lib.js
ds26gte: Done

src/web/js/trove/chart-lib.js
ds26gte: Done

src/web/js/trove/chart-lib.js
ds26gte: Rewrote and done

PR #560, opened 2024-11-18
bring chart tests up to date

ds26gte:
- can't just include color (or image-structs) anymore because tan misnomer
- box plot tests' gray values appear to have shifted, failing tests


ds26gte:
(This is a minimal PR that ensures the old chart tests are corrected. The box plots test, in particular, appears to have bitrotted for reasons that have nothing to do with the dot chart impl.)

PR #569, opened 2025-01-06
VMT/embedding API

jpolitz:
This is a draft pull request for the embedding API, which is used by VMT (https://vmt.mathematicalthinking.org/ allows the creation of “Pyret rooms”) and by the VScode plugin (https://marketplace.visualstudio.com/items?itemName=PyretProgrammingLanguage.pyret-parley)

I'm making it a draft because I'm not totally sure this is ready to merge, but want to get a look at it and have some conversation. It *should* have no change to any use of CPO-of-today, and just layer on the embedding API atop the same endpoint (`/editor`)

There are a few components to this:

- The embedding API works by assuming CPO has been iframed, and exposing various `postMessage` tools. It supports a number of actions, including initializing and updating the editor, triggering run events and interactions events, reading off the current state, and more. For the VScode plugin, it also exposes an RPC interface to expose the filesystem from the plugin to the embedded instance
  - Some examples of embedding tests are here: https://github.com/brownplt/code.pyret.org/tree/vmt-initialState/test-util/embed
  - The main logic for handling `postMessage`s is in `events.js`, a new file and concept for this set of features: https://github.com/brownplt/code.pyret.org/blob/vmt-initialState/src/web/js/events.js
- There is also a layer of instrumentation added to a number of places in the editor. For example, when run of definitions happens, it triggers an event that can be sent _out_ of the embedded instance to the enclosing page. This allows for environments like VMT to listen for events from the active editor and broadcast them to the other clients. (https://github.com/brownplt/code.pyret.org/blob/vmt-initialState/src/web/js/cpo-main.js#L394) There is a new concept and  logic around these events to detect whether they are “synthetic” or not. The issue here is that when, say, a run event is broadcast, we want the “controlled” clients to trigger the run, but then _not_ rebroadcast it/report it as an event. So there's some delicate stuff here, maybe some better patterns to follow, etc.

Comments welcome!


jpolitz:
Tagging @exidy80 (who really ought to be credited as a co-author on large chunks of this).

Also this is deployed at https://pyret-vmt-dfb765867402.herokuapp.com/, which is the endpoint that the VScode plugin and VMT use (https://github.com/jpolitz/pyret-parley-vscode/blob/5de96b396adfa41c6ab84c562e12b34f33763012/src/pyretCPOEditor.ts#L192)

src/web/js/authenticate-storage.js
```diff
@@ -5,6 +5,13 @@ window.storageAPI = storageAPIDeferred.promise;
 window.sheetsAPI = sheetsAPIDeferred.promise;
 
 window.handleClientLoad = function handleClientLoad(apiKey) {
+  debugger;
```
jpolitz: Remove stray debugger

src/server.js
```diff
@@ -773,70 +766,106 @@ function start(config, onServerReady) {
     return ret.promise;
   }
 
+    // NOTE(joe): this is a hack for making share URLs work on another server.
```
jpolitz, line 771: Lift this up into a discussion on the PR itself.

src/web/js/beforePyret.js
```diff
@@ -1294,19 +1297,37 @@ $(function() {
...
     CPO.documents.set("definitions://", CPO.editor.cm.getDoc());
   });
 
+  console.log("About to load Pyret: ", originalPageLoad, Date.now());
+
   var pyretLoad = document.createElement('script');
-  console.log(process.env.PYRET);
-  pyretLoad.src = process.env.PYRET;
+  console.log(window.PYRET);
+  pyretLoad.src = window.PYRET;
```
jpolitz, line 1330: Check if this is needed (was added for trying to embed as a webview directly in VScode, but that may not matter anymore because we are embedding a real built page)

jpolitz (commented):
Quick initial review highlighting a few things

.travis.yml
```diff
@@ -16,6 +16,7 @@ env:
   - PORT=5000
   - SAUCE_TEST_TARGET="http://localhost:5000"
   - SAUCE_USERNAME="pyret-lang"
+  - POSTMESSAGE_ORIGIN="*"
```
jpolitz, line 19: Why is this in .travis.yml? Is it to make tests work or something?

cpo-standalone.js
```diff
@@ -157,7 +157,9 @@ requirejs(["pyret-base/js/runtime", "pyret-base/js/post-load-hooks", "pyret-base
         $("#runDropdown").attr("disabled", false);
         clearInterval($("#loader").data("intervalID"));
         $("#loader").hide();
+        CPO.triggerOnLoad();
```
jpolitz: Maybe `CPO.events.trigger...`

package.json
```diff
@@ -14,6 +14,7 @@
     "body-parser": "^1.20.2",
     "bootstrap": "^4.6.2",
     "browserify": "^14.5.0",
+    "chromedriver": "^127.0.2",
```
jpolitz: This should probably be supplanted by the `.travis.yml` rules

src/server.js
```diff
@@ -773,70 +766,106 @@ function start(config, onServerReady) {
...
+    // The config variable here shouldn't be set to the same URL as the deployment
+    // URL (since this will just make self request). But since Google credentials
+    // lock in files to a particular server, and we want share links from CPO
+    // to work on our testing or other deployments, we need this.
+  function sharedFallback(url) {
```
jpolitz: sharedPrefetch/sharedfastPath or similar.

src/server.js
```diff
@@ -773,70 +766,106 @@ function start(config, onServerReady) {
...
-    contents.then(function(response) {
-      response
-        .on("response", (r) => r.headers["content-disposition"] = `inline; filename="${req.query.sharedImageId}"`)
-        .pipe(res);
-    })
-    .fail(function(err) {
-      res.status(400);
-      res.send("Could not access shared file, or shared file was not an image.");
-      res.end();
+    const requestURL = `${config.sharedFetchServer}/shared-image-contents?sharedProgramId=${req.query.sharedProgramId}`;
```
jpolitz: There's a bug on this line. Look up `sharedProgramId` vs `sharedImageId`

src/web/editor.html
```diff
@@ -4,6 +4,7 @@
   <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
   <title>code.pyret.org</title>
   <link rel="preload" href="{{&PYRET}}" as="script">
+  <script>window.PYRET = "{{&PYRET}}";</script>
```
jpolitz, line 7: I think this was just needed for webview embedding? Probably not needed now if we are iframe embedding.

src/web/js/authenticate-storage.js
```diff
@@ -5,6 +5,13 @@ window.storageAPI = storageAPIDeferred.promise;
...
+  if(!gapi || !gapi.client) {
+    storageAPIDeferred.reject("no gapi.client");
+    sheetsAPIDeferred.reject("no gapi.client");
+    console.log("Not logged in; proceeding without login info");
+    return;
+  }
```
jpolitz, line 13: Is this also only a webview thing?

src/web/js/before
jpolitz, line 1: Typo that this file got create

src/web/js/beforePyret.js
```diff
@@ -1391,6 +1412,14 @@ $(function() {
...
   }
 
+  const onLoadHandlers = [];
+  function onLoad(handler) {
+    onLoadHandlers.push(handler);
+  }
+  function triggerOnLoad() {
+    onLoadHandlers.forEach(h => h());
+  }
+
```
jpolitz, line 1412: These can be bundled up in an abstraction (and put on `.events` as suggested above)

src/web/js/beforePyret.js
```diff
@@ -1417,7 +1449,17 @@ $(function() {
...
+  let initialState = params["get"]["initialState"];
+
+  if (typeof acquireVsCodeApi === "function") {
+    window.MESSAGES = makeEvents({
+      CPO: CPO,
+      sendPort: acquireVsCodeApi(),
+      receivePort: window,
+      initialState
+    });
+  }
```
jpolitz, line 1445: Is this used at all in the iframed setting? Again check webview

src/web/js/events.js
```diff
@@ -1,115 +1,414 @@
-let counter = 0;
+let messageCounter = 0;
 let targetOrigin = POSTMESSAGE_ORIGIN;
```
jpolitz, line 2: Can this be `*` and what does it mean if it is?

src/web/js/events.js
```diff
@@ -1,115 +1,414 @@
...
+  let callbackCounter = 0;
+
+  function sendRpc(data, callback) {
+    callbackCounter += 1;
+    const callbackId = `${callbackCounter}-${data.module}-${data.method}`
+    callbacks[callbackId] = callback;
+    config.sendPort.postMessage({
+      protocol: "pyret-rpc",
+      data: { ...data, callbackId }
+    }, '*');
```
jpolitz, line 15: Should this be `'*'`?

src/web/js/events.js
jpolitz: (Does it *have* to be `*` to avoid really complicated issues with VScode webviews?)

src/web/js/events.js
```diff
@@ -1,115 +1,414 @@
...
+    });
+  });
+
+  const comm = commSetup(config, onmessage, gainControl, loseControl);
+
+  function editorUpdate(newCode) {
+    editor.cm.replaceRange(
+      newCode,
+      { line: 0, ch: 0},
+      { line: editor.cm.lastLine(), ch: 99999 },
```
jpolitz, line 183: This may break with image literals

src/web/js/events.js
```diff
@@ -1,115 +1,414 @@
...
+      replCM().refresh();
+    }, 100);
+  }
 
   // Thanks internet! https://github.com/codemirror/CodeMirror/issues/3691
   const thisAPI = "@ignore-this-api";
 
   editor.cm.on("change", function (instance, change) {
-    if (change.origin === thisAPI) {
+    if (change.origin === thisAPI || change.origin === "setValue") {
```
jpolitz, line 205: This may warrant a comment: it happens that CPO never does a setValue that we want to propagate to others. And it *does* do some setValues (of initial contexts and some other stuff) that we *don't* want to propagate)

src/web/js/events.js
```diff
@@ -1,115 +1,414 @@
...
+      thisAPI
+    );
+    editor.cm.refresh();
+  }
+
+  function replUpdate(newCode) {
+    replCM().replaceRange(
+      newCode,
+      { line: 0, ch: 0},
+      { line: replCM().lastLine(), ch: 99999 },
```
jpolitz, line 193: Can we call get length of last line/get last pos?

test/basic.js
```diff
@@ -18,6 +18,8 @@ describe("Basic page loads", function() {
   it("should load the editor", function(done) {
     this.timeout(80000);
     var self = this;
+    self.browser.executeScript("");
```
jpolitz: Why is this here?

.travis.yml
jpolitz: Oh, this makes sense. The `.travis.yml` section around this *is* the `.env` for Travis.

PR #571, opened 2025-01-09
set the title attribute of the filename element to be the FULL filename

schanzer:


jpolitz:
Is this intentionally just for blocks? There's a `beforeBlocks.js` and a `beforePyret.js` (it was more expedient to just copy when block started) so this small change would need to happen in two places.

(Coalescing these would be a good thing for me to work on soon)

schanzer:
*facepalm* 

Sorry about that. Done.

Issue #185, opened 2017-06-27
Table rendering in REPL does not have high enough contrast.


jswrenn:
The shading totally disappears when CPO is displayed on a projector.

schanzer:
If it doesn’t work for *sighted* users in classrooms, this feels more like a usability issue than an accessibility one.

If our default CSS fails to work in this setting, it means we need better CSS

schanzer:
You’re going to hate me for saying this, but all I can say is “we should hold ourselves to a higher standard.” We could alternate shadings between odd and even number rows. We could use wider borders, darker borders, or whatever — either way, I just can’t believe we can’t come up with something that works for projectors.

blerner:
Nonsequitur. We *do* alternate shading on each row, and we do have borders. Etc. What we want to avoid is that the backgrounds of the rows obscure the contents of the rows, especially when the rows contain images with arbitrary background colors themselves... So we're limited in what palettes we can use.

Imho, this is the same as the problem that PowerPoint color schemes look good on either your own screen or the projector screen, but almost never both :-\ I think it's probably a good idea to consider having two separate color schemes, as a result.

schanzer:
We alternate them already? Wow, it's barely visible on my own monitor (an admittedly-old Apple 27" cinema display).  The fact that I didn't realize this until now suggests that our current solution may not be as good as it could be.

I hear you on not wanting to hide the contents of the table, but as long as we have alternating rows, we run the risk of obscuring the table contents for two specific shades. Making one of those shades darker doesn't increase that risk, does it?

I think our use-case is _far_ simpler than PPTs, and that we should be able to do a better job here. Why not have the even rows just be slightly darker? Some simple experimenting in the Web Inspector gets the attached screenshot, and all I had to do was darken the even rows a few shades. While I was at it, I also made the header row much darker and more visually-distinct from data rows, which I think we need anyway. 

![image](https://user-images.githubusercontent.com/77126/42464602-90282d26-8377-11e8-9c18-327df2f95bcb.png)

schanzer:
FYI, if you like the screenshot above, this is just two CSS file changes:

editor.css line 1690: #e2e2e2
editor.css line 1664: #666

schanzer:
@jpolitz mind if I tackle this in a tiny PR (or naked commit)?

blerner:
Don't be _too_ tiny -- keep in mind you need to double-check all the themes we currently support.

PR #572, opened 2025-01-09
tweak CSS for even-numbered rows to make them more apparent (#185)

schanzer:
I only changed the default theme, and only very slightly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,7 +62,7 @@ before_deploy:
 install:
 - git submodule init
 - git submodule update
-- npm update # to update Pyret to the newest version
+- npm update pyret-lang # to update Pyret to the newest version
 - npm ci --ignore-scripts
 - make web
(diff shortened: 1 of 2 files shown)


blerner:
As a change, this is trivial and safe and can be merged.  Do you have a projector to try it on, to see if it suffices for your needs?  And, are there any corresponding changes needed in other themes?

schanzer:
Default is the only thing we use, and I wanted to make this as minimally invasive as possible. I don't have a projector to try it on, but if you're comfortable I can push another commit that makes the _other_ rows darker by a similar amount. That should make the distinction quite noticeable, even on projectors

jpolitz:
Thanks! There was a build issue on horizon (just introduced by an external change to a dependency, but our fault for npm updating in the middle of the build script). I'm pulling that change into PRs so they will build before merging.

Issue #573, opened 2025-01-16
Rules for whitespace in Pyret strings are ignored when displayed in an SVG


schanzer:
If a reactor displays an SVG element (as is the case for literally every chart) and the SVG element has a TEXT element within it, that string will be displayed using the browser's default whitespace-collapsing behavior.

This is a problem for charts of ours that include a lot of data in their titles:

![Image](https://github.com/user-attachments/assets/0456cfbb-389a-4109-ab7b-86c8ec02b82c)

We'd like to separate those strings, but adding spaces between them in Pyret doesn't help because of this issue.

PR #574, opened 2025-01-16
preserve whitespace in svg text elements, when used in a reactor window

schanzer:
(addresses #573)

### DIFF
--- a/src/web/css/editor.css
+++ b/src/web/css/editor.css
@@ -1583,7 +1583,8 @@ code div {
 .ui-widget-content{
   background: none;
 }
-.ui-widget-content svg text { white-space: pre; }
+
+.whitespace-pre { white-space: pre; }
 
 /* From internet: http://stackoverflow.com/questions/5663137/jquery-modal-dialog-is-hidden-by-overlay-div-if-dialogs-parent-has-fixed-positi */
 .ui-widget-overlay{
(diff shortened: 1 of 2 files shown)


blerner:
This seems like the wrong fix to me, or at least, is too stringent.  I suspect you want `white-space: pre-wrap`, since you want text to look like well-formatted text, not necessarily code.  And what you likely really wanted was multiple lines of text, so you want to the white-space option that preserves linebreaks but in other respects formats text as text, hence `pre-wrap`.

schanzer:
I want a string to render as a `TEXT` node within an SVG exactly as it would render in Pyret using `text`. No line wrapping. (although it would be pretty cool if Pyret supported line wrapping someday!)

schanzer:
@blerner unfortunately, making this selector more restrictive is quite difficult!

Google Charts has total control over the DOM structure of a chart, and doesn't provide any helpful nodes, classes or ids that would let us identify _this specific text_. The best we can do is walk the DOM:

`.ui-widget-content svg > g:nth-child(2) > text { white-space: pre; }`

This works for our use-case, and from spot-checking all of our charts this causes no problems. But if Google ever changes their DOM tree, this will be yet another thing that breaks.

If this doesn't leave you feeling good about the PR, close it -- our userland workaround is good enough.

blerner:
When you generate labels, are they _strings_, or can they be _DOM nodes_?  If the latter, then generate `<span class="our-specific-tag">{the string label}</span>`, so you can trigger off of that...

schanzer:
They're just strings. Google's API gives us only very primitive options

PR #580, opened 2025-02-14
add pinhole support to place-image-align, and change add-line to work from the top-left corner

blerner:
We don't yet have any test cases that notice this change; I ran it manually on one example so far.  I'll try to add some more test cases soon.

### DIFF
--- a/src/web/js/trove/make-image.js
+++ b/src/web/js/trove/make-image.js
@@ -620,10 +620,12 @@
         var placeX = unwrapPlaceX(maybePlaceX);
         var placeY = unwrapPlaceY(maybePlaceY);
         var background = unwrapImageOrScene(maybeBackground);
-        if      (placeX == "left"  ) { x = x + img.getWidth()/2; }
-        else if (placeX == "right" ) { x = x - img.getWidth()/2; }
-        if      (placeY == "top"   ) { y = y + img.getHeight()/2; }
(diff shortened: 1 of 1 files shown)


blerner:
@dbp is this obsoleted by your https://github.com/brownplt/pyret-lang/pull/1802 ?  I think the `place-image-align` part is still useful, but your change for `add-line` is different from this one...

schanzer:
@dbp checking in about Ben's comment here.

blerner:
I think the pinhole support is missing from the other PR? But regardless, this particular CPO PR is obsolete because the image library has moved to pyret-lang. @dbp can you confirm where both parts of your improvements are, and whether we've fixed the crashing bug in your other PR? And do we have tests for this so we can merge it?

dbp:
IIRC, when I changed `add-line`, I didn't change some other functions that were nearby, that seemed like maybe they could use refactoring, as they were implemented very differently (but, one upside meant that they didn't have the bug that `add-line` had). As far as I know, the bug that @jpolitz saw for command-line node-canvas still exists (unless he fixed it)? 

I did not add automated test cases in that PR (not sure the workflow for doing that with image stuff? Do existing image tests test pixel output? Or compare for equality for different operations? Or?). 

Not quite sure what I'm being asked here -- I do _not_ know this code well :)

PR #585, opened 2025-02-26
Progress on an all-local setup for built CPO

jpolitz:
Design goals:

- Make it so CPO runs with the network on your device *turned off*
- Clearly identify all the dependencies we need for a running CPO instance
- More easily support (primarily for a VScode extension-like context) an all-local built bundle. Right now in the VScode extension, you _cannot_ load a new Pyret editor if your WiFi cuts out. Having this all-local would make it much more straightforward to run in that kind of context (all-offline), and would also be good for privacy (knowing no requests go out).

This initial commit does a few things:

1. Downloads relevant 3rd party JavaScript (fonts, and JQuery stuff) and copies it in Makefile
2. Does some housekeeping on different failure cases to make the page load even if e.g. Google scripts don't load

After building, you can try this out for now by setting `PYRET` to e.g. `/js/cpo-main.jarr`, then

```
node make-template src/web/editor.hml > build/web/editor.html
cd build/web/
python3 -m http.server 8080
// turn off your network and open localhost:8080/editor.html
```

Note that right now the *.html files are *dynamically* templated, so the extra build step (which maybe should unconditionally happen) creates the correctly-filled-in editor.html

Some things simply cannot be made to work in this way. The chart loader and google API loader I left alone; even if you download those files and load them from source locally they immediately make network requests, embed a 3rd party iframe, etc. So it's unlikely that stuff will ever work offline.

It's an open (and pretty important) question if the Google Chart library has a good offline build mode so that e.g. you can connect CSV-file charts to that library


blerner:
FYI: https://developers.google.com/chart/interactive/faq
> Can I use charts offline?
    Your users' computers must have access to https://www.gstatic.com/charts/loader.js in order to use the interactive features of Google Charts. This is because the visualization libraries that your page requires are loaded dynamically before you use them. The code for loading the appropriate library is part of the included script, and is called when you invoke the google.charts.load() method. Our [terms of service](https://developers.google.com/chart/terms) do not allow you to download the google.charts.load or google.visualization code to use offline.
Can I download and host the chart code locally, or on an intranet?
    Sorry; our [terms of service](https://developers.google.com/chart/terms) do not allow you to download and save or host the google.charts.load or google.visualization code. However, if you don't need the interactivity of Google Charts, you can screenshot the charts and use them as you wish.

Issue #556, opened 2024-10-17
Error loading a table from a gsheet


schanzer:
Joy reported this from a teacher...

Run [this program](https://code.pyret.org/editor#share=1WqFQ_uZtplDvcbi35ojDWZzOVbREP4un&v=dc8be4e), then evaluate `Cell-TABLE` in the interactions area. You can load the spreadsheet linked from the Pyret file to check out what's going on.

It looks like Pyret's importing-from-google-sheets functionality is getting tripped up by extremely small values in google sheets.

blerner:
The table is

Molecules | Species | Speed
-- | -- | --
Na+ | ions | 5.00E-04
K+ | ions | 4.70E-04
O2 | small molecules | 2.30E+12
CO2 | small molecules | 3.50E+08
H2O | small molecules | 3.40E+04
EtOH | small molecules | 2.10E+04
Steroids | small molecules | 1.00E+04
Urea | small molecules | 5.40E-02
Glycerol | small molecules | 5.40E-02
small molecule drugs | small molecules | 1.00E+00
Cyclosporin | peptides | 2.50E-04
TAT | peptides | 2.70E+04



And Pyret produces 

Molecules | Species | Speed
-- | -- | --
"Na+" | "ions" | 0.0005
"K+" | "ions" | 0.00047
"O2" | "small molecules" | 2300000000000
"CO2" | "small molecules" | 350000000
"H2O" | "small molecules" | 34000
"EtOH" | "small molecules" | 21000
"Steroids" | "small molecules" | 10000
"Urea" | "small molecules" | 0.054
"Glycerol" | "small molecules" | 0.054
"small molecule drugs" | "small molecules" | 1
"Cyclosporin" | "peptides" | 0.00025
"TAT" | "peptides" | 27000

Which of those values look wrong to you?

schanzer:
@blerner Are you able to run the program I linked, and evaluate `Cell-Table`? On my machine (up-to-date Brave, macOS, plenty of RAM) it dies with a cryptic "error displaying value"

blerner:
I ran it and literally copied and pasted cpo's output. I see no problem on my machine.  It runs in both Firefox and Chrome.

schanzer:
@blerner oh crap - looks like the teacher tweaked the numbers to avoid the bug. Sorry, I didn't think they'd touch the file!

[...]

PR #588, opened 2025-03-25
Added missing 3rd (of 4) arg to jsnums.toRepeatingDecimal() calls #556

ds26gte:
These occur in ide.js and output-ui.js.

### DIFF
--- a/src/web/js/ide.js
+++ b/src/web/js/ide.js
@@ -135,7 +135,7 @@ function makeRuntimeAPI(CPOIDEHooks) {
       // This function returns three string values, numerals to
       // appear before the decimal point, numerals to appear
       // after, and numerals to be repeated.
-      var decimal = jsnums.toRepeatingDecimal(num.numerator(), num.denominator(), runtime.NumberErrbacks);
+      var decimal = jsnums.toRepeatingDecimal(num.numerator(), num.denominator(), undefined, runtime.NumberErrbacks);
       return {
         type: 'number',
         value: {
--- a/src/web/js/output-ui.js
+++ b/src/web/js/output-ui.js
@@ -1399,7 +1399,7 @@
           // after, and numerals to be repeated.
           var numr = num.numerator();
           var denr = num.denominator();
-          var decimal = jsnums.toRepeatingDecimal(numr, denr, runtime.NumberErrbacks);
+          var decimal = jsnums.toRepeatingDecimal(numr, denr, undefined, runtime.NumberErrbacks);
           var prePointString = decimal[0];
           var postPointString = decimal[1];
           var decRpt = decimal[2];


blerner:
The code change looks reasonable. But do we have a test to make sure this is correct? The original issue definitely had something that triggered this bug, so we should be able to extract a regression test for this...

schanzer:
@blerner @ds26gte [here's the file that exposes the bug](https://code.pyret.org/editor#share=1WqFQ_uZtplDvcbi35ojDWZzOVbREP4un&v=dc8be4e) (evaluate `Cell-TABLE` after running). [Here's the offending spreadsheet](https://docs.google.com/spreadsheets/d/1MDCFhQl2tw-FdALuBS8vXC3Kom11aMocgD1OPVS256c/edit?gid=0#gid=0).

jpolitz:
This seems like much more than a missing errback issue: somehow the parsing of this sheet is leading to a rational with a 0 denominator. The string renderers show it, the `toRepeatingDecimal` call in the DOM renderer errors:

image

https://code.pyret.org/editor#share=1a1Pd-fdFlF6rj3QfkDP-St8LNbhzdV8r&v=fee2ecd

The data is:


Molecules | Species | Speed
-- | -- | --
Na+ | ions | 5.00E-19

So somehow in the sheets reader we are getting `1298074214633707/0` as a value. In the debugger that value looks like this:

image

jpolitz:
Sheet URL: https://docs.google.com/spreadsheets/d/1MDCFhQl2tw-FdALuBS8vXC3Kom11aMocgD1OPVS256c/edit?gid=0#gid=0

jpolitz:
Note that the effect of the code in this PR is just to raise a nicely rendered `DomainError`, which is good, but should _not_ happen on this input.

blerner:
This sounds like something that's easily reproducible as a JS regression test, then?  I.e. take the string `5E-19` (or `5.00E-19`) and parse it and make sure we don't get a broken number?

jpolitz:
The constant in a Pyret program is fine. It's something in the sheets reading/parsing.

image

blerner:
But I think the field content is _actually_ `5E-19`, without the `.00` -- I suspect the cell formatting is showing two decimal places.  And in CPO,

image

According to Pyret, the value inside that `some` has a `to-repr` of `5e-19`, a `to-string` of `5e-19`, and yet renders as _nothing at all_ in the output...  So something is very strange with that value.

jpolitz:
What browser are you on?

Brave Version 1.76.82 Chromium: 134.0.6998.178 (Official Build) (arm64)

image

blerner:
Hmmm, I may have had stale state in my browser window somehow; refreshing made the problem go away, so /shrug?

jpolitz:
OK, the culprit is `fromFixnum` in js-numbers, which is somehow wrong (IDK how yet) in the logic that was added for handling scientific notation literals:

https://github.com/brownplt/pyret-lang/commit/80477929399e76d5939230bc56834b9bdf388750#diff-34cd626ab11424687bfd071b24342ee9d60c879050d69239ca8cf86df95e2f39R7-R77

`fromFixnum` is relatively hard to reach from Pyret (most things use `makeNumberFromString` which uses `jsnums.fromString`). However, `json` uses it, so a simple command-line repro is:

```
→ cat json-5e.arr
import json as J

print(J.read-json("5E-19"))
→ make json-5e.jarr
→ node json-5e.jarr
j-num(1298074214633707/0)
```

jpolitz:
@ds26gte if you're still on this: a useful way to proceed would be to figure out how to write tests for `fromFixnum` (which likely was never tested very thoroughly), perhaps using `json`, and then make sure it works for many cases with `E` literals.

ds26gte:
PR brownplt/pyret-lang#1810 addresses the fact that the error shouldn't happen (as opposed to having a decent error message).

schanzer:
@jpolitz now that Dorai's PR (mentioned above) is merged into horizon, I can confirm that this issue is resolved. See https://pyret-horizon.herokuapp.com/editor#share=1WqFQ_uZtplDvcbi35ojDWZzOVbREP4un&v=dc8be4e.

@ds26gte if this means your other PR subsumes this one, please close. Otherwise, let us know what unique fixes are in this one?

blerner:
These fixes are needed also, and are not subsumed by another PR as far as I know. They also are not tested, and I'm not thrilled about that...

Also, while tracing through this code into js-numbers, I found *another* missing errbacks argument, in the call to equals() on line 3905.

ds26gte:
PR https://github.com/brownplt/pyret-lang/pull/1810 (accepted) took care of the fact that this bug shouldn't happen.

PR https://github.com/brownplt/pyret-lang/pull/1818 (pending) ensures
- proper error message on invalid args to `toRepeatingDecimal()`, with JS-tests (the only tests possible now, now that the bug can't happen via Pyret commandline or CPO)
- all calls to `equals()`, `lessThan()`, `lessThanOrEqual()` take errbacks. (`greaterThan()` and `greaterThanOrEqual()` were already clean)

schanzer:
@ds26gte are you saying that this PR should be merged _after_ #1818? Or that it's obviated by #1818 and should be closed?

ds26gte:
> @ds26gte are you saying that this PR should be merged _after_ #1818? Or that it's obviated by #1818 and should be closed?

It could have been merged any time in the past and now. It fixes a call to `toRepeatingDecimal()` in CPO that was using a bad set of arguments, where an errbacks argument wasn't really an errbacks.  

PR #1818 did not change the argument signature of `toRepeatingDecimal()` at all, so the fix introduced by this PR is still valid. However, an entirely different bug that had to do with misreading VERYSMALL rational numbers was fixed via PR #1810, so this bug is no longer exercised.

PR #1818 has added tests that ensure that if the original bug were still unfixed, a proper error message would have been raised. It does this by simulating (via JS) a situation where the original bug could still happen.

ds26gte:
P.S. If you're asking me if this PR should be merged or abandoned, it should be merged!

schanzer:
@blerner does [#1818](https://github.com/brownplt/pyret-lang/pull/1818) add the tests you wanted, to allow this PR to be merged?

blerner:
I'm not thrilled with the behavior of `undefined-- <= 0`: I had to look that one up, twice, to remind myself what was supposed to happen. (Undefined turns to NaN, which isn't comparable with anything -- two subtle type coercions). Could you add a line at the top of toRepeatingDecimal that says `limit = limit ?? Infinity;`, to make the behavior clearer?

blerner:
I don't see any regression tests in #1818 that confirm that the missing argument would cause a problem. If I'm reading correctly, you've tested that toRepeatingDecimal *when given four validly typed arguments* will throw an errbacks error. A regression test for this ought to test that omitting the third (optional) argument causes a method-not-found js error, I think...?

Or, better, since errbacks is now a required argument, we should make limit be a required argument, and replace undefined with +Infinity. Then it's a trivial check for `if (typeof limit !== "number")`, to throw a contract error as needed. (We shouldn't have an "optional" argument before a required argument, and errbacks needs to be required.)

ds26gte:
> I'm not thrilled with the behavior of `undefined-- <= 0`: I had to look that one up, twice, to remind myself what was supposed to happen. (Undefined turns to NaN, which isn't comparable with anything -- two subtle type coercions). Could you add a line at the top of toRepeatingDecimal that says `limit = limit ?? Infinity;`, to make the behavior clearer?

This — i.e., relying an JS comparison quirk — is not what's happening though.

`toRepeatingDecimal()` does not directly take a `limit` argument at all. It rather takes an `options` argument, which can legitimately be `undefined`, which is indeed taken care of correctly.  In order to specify a non-default limit, this options should be an object with field `limit`.  Otherwise, the limit is set to 512, a very valid number. It is never `undefined`, and there is no reliance on `undefined-- <= 0`. 

Using `Infinity` as an explicit 3rd argument to `toRepeatingDecimal()` would actually be a type error.

ds26gte:
> I don't see any regression tests in #1818 that confirm that the missing argument would cause a problem. If I'm reading correctly, you've tested that toRepeatingDecimal _when given four validly typed arguments_ will throw an errbacks error. A regression test for this ought to test that omitting the third (optional) argument causes a method-not-found js error, I think...?
> 
> Or, better, since errbacks is now a required argument, we should make limit be a required argument, and replace undefined with +Infinity. Then it's a trivial check for `if (typeof limit !== "number")`, to throw a contract error as needed. (We shouldn't have an "optional" argument before a required argument, and errbacks needs to be required.)

Commit brownplt/pyret-lang@1ec7a95 adds a test for when `toRepeatingDecimal()` is called with errbacks as a 3rd rather than 4th argument.

For 3rd argument being explicitly +Infinity, please see other comment for why this, rather than undefined, would be a type error.

blerner:
Oh... Ok, so my problem then was that the docstring comment was misleading, but not inaccurate: 

```
   // An optional limit on the decimal expansion can be provided, in which
   // case the search cuts off if we go past the limit.
   // If this happens, the third argument returned becomes '...' to indicate
   // that the search was prematurely cut off.
   var toRepeatingDecimal = (function() {
     var getResidue = function(r, d, limit, errbacks) {
```

That proximity sure makes it look like the third argument to this function is "an optional limit"!
It might be helpful just to add a comment above getResidue that explains that limit is determined from options, below. Or better yet, extract getResidue out of this function, since it doesn't need to be closed over within toRepeatingDecimal, and since it's complicated enough that it deserves its own explanation.

ds26gte:
> Oh... Ok, so my problem then was that the docstring comment was misleading, but not inaccurate:
> 
> ```
>    // An optional limit on the decimal expansion can be provided, in which
>    // case the search cuts off if we go past the limit.
>    // If this happens, the third argument returned becomes '...' to indicate
>    // that the search was prematurely cut off.
>    var toRepeatingDecimal = (function() {
>      var getResidue = function(r, d, limit, errbacks) {
> ```
> 
> That proximity sure makes it look like the third argument to this function is "an optional limit"! It might be helpful just to add a comment above getResidue that explains that limit is determined from options, below. Or better yet, extract getResidue out of this function, since it doesn't need to be closed over within toRepeatingDecimal, and since it's complicated enough that it deserves its own explanation.

Fixed in commit brownplt/pyret-lang@7b63e0f

- `getResidue()` definition moved out of `toRepeatingDecimal()`
- both functions now have better comments
- jsnums-test.js now tests both functions for a too-small limit

ds26gte:
I'll summarize the situation so we can devise a proper regression test for this in CPO.

The bug that this fix addresses is incorrect errbacks propagation.

The pyret-lang function `toRepeatingDecimal()` raises an error when
- its first arg (numerator) is not an integer
- its second arg (denominator) is not an integer
- its second arg is 0

None of these situations should ever happen, because `toRepeatingDecimal()` is called with the numerator and denominator of a rational number, so n and d are indeed integers (or bigintegers) and d is never 0.

However, pyret-lang's `fromFixnum()` had a serious bug and for very-small fixnums it created faulty rationals. E.g., 1e-7 was being read as

``` 
Rational { 
  n: 944473296573929, 
  d: BigInteger { t: 0, s: 0 } 
} 
```

This is clearly wildly wrong: `n` is wrong, and `d` is not only wrong but 0. I fixed this in brownplt/pyret-lang#1810. In this case, the more serious error is the denominator being a bigint version of 0. Assuming `fromFixnum()` wasn't fixed, the CPO program

``` 
import json as J 
J.read-json('1e-7') 
```

should produce a `d equals 0` error.

However, since CPO calls `toRepeatingDecimal()` with 3, rather than 4, args, the errbacks arg as seen by `toRepeatingDecimal()` is `undefined`. I.e., errbacks is incorrectly propagated. Therefore the expected error message isn't produced.  Instead we get

``` 
<error display value: details logged to console> 
```

By fixing the call in CPO, we _would_ get the correct error message.

(Except that, since PR 1810 has been accepted into pyret-lang, the error no longer happens.)

We now have exhaustive tests in pyret-lang that show

1. `fromFixnum()` correctly reads very-small fixnums like 1e-7 and 5e-19
2. `toRepeatingDecimal()` produces the _correct_ error when given incorrect arguments (for all three kinds of incorrectness) and a _correct_ errbacks
3. `toRepeatingDecimal()` produces an _incorrect_ error when given incorrect args (for all three kinds of incorrectness) and an _incorrect_ errbacks

It remains to add similar tests in CPO, to show that correcting the call `toRepeatingDecimal()` in CPO doesn't introduce any CPO bugs. Ideas welcome.

jpolitz:
For what it's worth, this arity mismatch has been around for about 9 years:

- The errbacks parameter was added in the library here: https://github.com/brownplt/pyret-lang/commit/1fe46142430bc2c70abc55bf613973282c260e1f
- The errback was added to CPO here: https://github.com/brownplt/code.pyret.org/commit/c00ef93aa7ae624e0496de209fba10998a04c7bb

This is clearly a case of optional trailing arguments being harmful, because the natural “add the default errbacks” refactoring was wrong, because this function call had elided the last argument before the errbacks argument was added.

For various reasons of how the `options` argument is interpreted, this didn't cause any bad visible behavior.

In fact, this line of code _has never caused a user visible bug_ and the change in this PR _is not something we can observe_. It will happen to work both ways as far as I can tell.

This entire thread is talking about a completely different issue with toFixnum that is completely unaffected by this code change.

Here is what a test looks like that makes sure the CPO code paths that render a repeating decimal are visited:

https://github.com/brownplt/code.pyret.org/commit/cad4c2ae3d8da1f5603a88c5959a6c9835221b9e

schanzer:
@ds26gte , please read the test case _carefully_ and make sure you understand how it works. I suspect that you will be writing similar ones of your own in the future, so consider this an opportunity to get a jump on it.

@jpolitz thanks for the summary - and the test case! Do you feel this PR is ready to merge, or would you like to see Dorai add any additional tests?

PR #589, opened 2025-04-10
Stylize tildes in roughnums to be more readable

blerner:
Requires https://github.com/brownplt/pyret-codemirror-mode/pull/12


blerner:
Here's what that looks like:
```
1.5
~1.5
-1.5
~-1.5
```
image

shriram:
This looks great. Let's roll with this and see how it fares when I teach with it in the fall. Thanks @blerner

PR #596, opened 2025-08-09
provide pie-chart, bar-chart, dot-plot, labeled-dot-plot, E

ds26gte:


src/web/arr/trove/dcic2024.arr
```diff
@@ -242,3 +247,5 @@ fun count-many(tab, cols):
       end)
   end
 end
+
+E = ~2.718281828459045
```
jpolitz: I think this is defined in `constants`, right?

ds26gte: Yes, however `dcic2024.arr` includes `essentials2024.arr`, which already sources `constants.arr` but deliberately only takes `PI`. 

Ideally we should change `essentials2024.arr` to also take `E` from `constants.arr`, but I didn't want to disturb anything there, especially if there is user code that uses `E` as a regular variable (i.e., without `shadow`).

jpolitz: But in this context can't you just `import constants as C` and then `provide from C: E end`?

PR #600, opened 2025-09-17
disable undo keyboard shortcuts if in VScode mode

jpolitz:
VScode has its own definition of undo/redo that send direct updates to the `TextDocument` backing the editor.

If CodeMirror _also_ listens to these, then they fight a bit; for example, the `TextDocument` updates, which tries to propagate its changes to the CodeMirror instance, which has just processed an undo, which then sends a message to update the `TextDocument`. This is especially bad for updates that _undo to re-add text_, because it can get interpreted as a double-insertion.

So, let VScode be in charge of these events only, and respect the `TextDocument` as the model.


jpolitz:
Closing because this code already made it into horizon by other means, and what's there seems to work just fine.

https://github.com/brownplt/code.pyret.org/commit/81097b35cd21089974671e84b2335383200cc3e7

PR #604, opened 2025-09-29
essentials2024.arr: provide string-{get-index,find{,-index,-opt}} from global

ds26gte:
This PR has `essentials2024.arr` `provide` the functions
```
string-find
string-find-opt
string-find-index
string-get-index
```
from `global`. (This is similar to how `starter2024.arr` `provide`s these functions.)

This has the effect of making these functions available from `dcic2024.arr` too.

This is a fix for the `pyret-lang` issue brownplt/pyret-lang#1824.


jpolitz:
I don't think we should change existing contexts for this reason. There might be a bunch of people who *already* added these definitions to their code as abbreviations, etc. We would break all their code.

ds26gte:
> I don't think we should change existing contexts for this reason. There might be a bunch of people who _already_ added these definitions to their code as abbreviations, etc. We would break all their code.

Oh, I was going by your statement that "it's reasonable to add string-find to both", where I thought the "both" referred to starter2024 and essentials2024...

jpolitz:
Yeah I should have said “we should have added string-find to both” but I think that train has sailed. We will make a new context for next year, though, and can add it then.

ds26gte:
A new PR #605, one that doesn't modify frozen context `essentials2024`, supersedes this one.

ds26gte:
Deleting PR.

PR #605, opened 2025-09-29
dcic2024.arr: provide string-{get-index,find{,-index,-opt}} from global

ds26gte:
This PR has `dcic2024.arr` provide the functions
```
string-find
string-find-opt
string-find-index
string-get-index
```
from `global`. 

This is a fix for the pyret-lang issue https://github.com/brownplt/pyret-lang/issues/1824.


schanzer:
@shriram confirming that you DO want this change, but pushed to a new context ("dcic2025" or similar)?

PR #622, opened 2026-03-26
make sure rotate elts in the table (like roughnums!) don't punch through

schanzer:


blerner:
This works, and it's 95% complete -- there's a _teeeny_ visual issue remaining, with the _border_ of a scrolled `thead` element:
image

The side borders scroll up, and the top border scrolls away (because it's attached to the _table_, not the _thead_), leaving a 1px gap of weirdness...

schanzer:
@blerner that issue is actually unrelated to the rotation thing -- it's just a part of using position:sticky. Anyway, I've fixed it in the latest commit.

blerner:
Probably good enough :)  there's still a silly little CSS glitch of the rounded-corners vs the side borders when scrolled, but I suspect it would take way too much time to finesse:
image

The right solution would be to use https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Conditional_rules/Container_scroll-state_queries, and the CSS from https://stackoverflow.com/a/57991537, and use that to unround the corners... but Firefox doesn't support scroll-state queries yet :(

PR #627, opened 2026-05-23
first stab at templating app_domain and app_name strings

schanzer:
Trying to template out strings that are user-facing vs. code-facing. This would make diffs with pyret.bootstrapworld.org a little easier, assuming I have the semantics right


.github/workflows/makefile.yml
```diff
@@ -27,28 +27,28 @@ jobs:
...
-        # Set Chrome binary path for tests
-        echo "GOOGLE_CHROME_BINARY=./chrome-linux64/chrome" >> $GITHUB_ENV
+        unzip chromedriver-linux64.zip
+
+        # Ensure binaries are executable
+        chmod +x chrome-linux64/chrome chromedriver-linux64/chromedriver
+
+        # Set absolute paths for tests (relative paths break when ChromeDriver spawns Chrome)
+        echo "GOOGLE_CHROME_BINARY=$GITHUB_WORKSPACE/chrome-linux64/chrome" >> $GITHUB_ENV
+        echo "CHROMEDRIVER_BINARY=$GITHUB_WORKSPACE/chromedriver-linux64/chromedriver" >> $GITHUB_ENV
```
jpolitz, line 43: I'm pretty sure this is breaking the CI.

What was failing? Like what's the “Claude says”? This dynamic “fetch-chromedriver” has been working for years.

schanzer: Here's Claude's reasoning behind the fix (copy/pasted):

--

1. Relative → absolute path: ./chrome-linux64/chrome → $GITHUB_WORKSPACE/chrome-linux64/chrome. ChromeDriver spawns Chrome as a subprocess and the working directory isn't guaranteed to be the repo root.
3. ChromeDriver extracted and wired up directly: Now we unzip chromedriver-linux64.zip and set CHROMEDRIVER_BINARY so util.js uses chrome.setDefaultService() with the exact binary, bypassing the npm chromedriver package entirely (and those two npm install chromedriver steps that were adding noise).
4. --disable-dev-shm-usage added: Chrome in Docker/CI containers shares /dev/shm by default (64MB limit) which is too small and causes crashes. This flag disables that.

jpolitz:
Thanks! I'm trying to fix the CI directly on horizon. I reverted the CI fix that's here.

src/web/editor.html
```diff
@@ -2,9 +2,9 @@
 <html>
 <head>
   <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
-  <title>code.pyret.org</title>
+  <title>{{APP_NAME}}</title>
   <link crossorigin="anonymous" rel="preload" href="{{&PYRET}}" as="script">
-  <script>window.PYRET = "{{&PYRET}}";</script>
+  <script>window.PYRET = "{{&PYRET}}"; window.APP_NAME = "{{APP_NAME}}"; window.APP_DOMAIN = "{{APP_DOMAIN}}";</script>
```
jpolitz, line 7: Do we need the `&` here? Any escaping that we should include/avoid? Should probably justify the difference.

src/web/faq.html
```diff
@@ -2,7 +2,7 @@
 <html>
 <head>
   <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
-  <title>code.pyret.org</title>
+  <title>{{APP_NAME}}</title>
```
jpolitz, line 5: Dumb question: does faq get templated by server.js? I forget which pages actually get all the template variables. Is it all of them?

src/web/editor.html
schanzer: oooh, good call.

src/web/faq.html
schanzer: Looks like it doesn't:
`app.get("/close.html", function(_, res) { res.render("close.html"); })`

Is there any reason not to pass `defaultOpts` to all of them?

